Issue #11, opened 2022-08-12
Fetch names of tokens


evan-gray:
It's unlikely that most people have memorized the token addresses.

PR #41, opened 2022-09-07
Issue #11 & added links for txhash & token addys

ckeun:


src/components/Governor.tsx
```diff
@@ -181,7 +253,12 @@ function Governor() {
   const [tokenSorting, setTokenSorting] = useState<SortingState>([]);
   const tokenTable = useReactTable({
     columns: tokenColumns,
-    data: governorInfo.tokens,
+    data: governorInfo.tokens.map((tk) => ({
```
evan-gray, line 256: I think this may cause the table to re-render every time because data will always be a new array - this can be avoided with a useMemo that creates this array when the dependencies of governorInfo.tokens and tokenSymbols change.

evan-gray: Better yet, you can use your useSymbolInfo hook to combine the info for you.

src/hooks/useSymbolInfo.ts
```diff
@@ -0,0 +1,418 @@
...
+  return;
+}
+
+const TIMEOUT = 60 * 1000;
+
+function useSymbolInfo() {
+  const { currentNetwork } = useNetworkContext();
+  useEffect(() => {
+    (async () => {
+      const response = await getGovernorTokenList(currentNetwork);
```
evan-gray, line 410: I think this will now get called twice, you can pass the result of this call in as a parameter and avoid the double call - this input array can serve as the list of tokens you need, and you can then build a return based on it.

src/hooks/useSymbolInfo.ts
```diff
@@ -0,0 +1,418 @@
...
+      await getTokenMetaData(response.entries);
+      await new Promise((resolve) => setTimeout(resolve, TIMEOUT));
+    })();
+  }, [currentNetwork]);
+  return TOKEN_CACHE;
```
evan-gray, line 415: Solely relying on data that is not returned from a hook won't cause this component to re-render - (this may appear to work now if re-renders are always being triggered). Using a cache is good, but within this component it will likely need to rely on a useState to update the value to be returned.

src/utils/consts.ts
```diff
@@ -0,0 +1,216 @@
...
+    platform: "solana",
+    covalentChain: 1399811149,
+    explorerStem: `https://solscan.io`,
+  },
+  2: {
+    name: "eth",
+    evm: true,
+    chainId: CHAIN_ID_ETH,
+    endpointUrl:
+      process.env.REACT_APP_ETH_RPC || "https://main-rpc.linkpool.io",
```
evan-gray, line 53: FYI, other examples have used ankr endpoints with some success.

evan-gray (changes requested):
Great changes, just some small render comments

evan-gray (approved):

PR #49, opened 2022-11-23
Added Mainnet network that uses the cloud functions for guardian data

kev1n-peters:


evan-gray (approved):
Sweet! Let's give it a go!

PR #81, opened 2023-04-06
cloud_functions/fly/database: Added TVL cloud functions, fly writes message counts to firestore

kev1n-peters:


evan-gray (approved):
Approving based on soak testing

PR #82, opened 2023-04-11
Acct: Add governed flag

bruce-riley:


dashboard/src/components/EnqueuedInGovChecker.tsx
```diff
@@ -0,0 +1,76 @@
...
+function EnqueuedInGovChecker({
+  transferKey: { emitter_chain, emitter_address, sequence },
+}: {
+  transferKey: PendingTransferKey;
+}) {
+  const { currentNetwork } = useNetworkContext();
+  const [vaaIsEnqueued, setVaaIsEnqueued] = useState<boolean | null>(null);
+  useEffect(() => {
+    (async () => {
+      setVaaIsEnqueued(null);
```
evan-gray, line 17: this can be moved outside the async

dashboard/src/components/EnqueuedInGovChecker.tsx
```diff
@@ -0,0 +1,76 @@
...
+      let messageId: MessageID = {emitterChain: emitter_chain, emitterAddress: emitter_address, sequence: sequence.toString() };
+      const isEnqueued = await isEnqueuedInGovernor(currentNetwork.endpoint, messageId);
+      setVaaIsEnqueued(isEnqueued);
```
evan-gray, line 20: when setting state in an async, always use a cancellable, otherwise there's a potential for your component to get destroyed and then call set state on a component that no longer exists. many other of the async hooks in this codebase use that pattern.

evan-gray (changes requested):
This is pretty good. The only downside to the component making the axios call is that when there are 4 items enqueued there end up being like 5 calls to `governor-stats` on the whole page. I think it might be best to hoist `const governorInfo = useCloudGovernorInfo();` to a shared component between `MainnetGovernor` and `Accountant` (`Main.tsx` could suffice for now). If it is in main, we would need to ensure that the hook doesn't fire off calls when not in mainnet. Alternatively, make a 'MainnetOnlyStuff' component that has everything from that ternary in `Main`

bruce-riley:
Reworked it to avoid the extra `useCloudGovernorInfo` calls.

evan-gray (approved):

PR #86, opened 2023-04-25
dashboard: update quorum calculation for testnet

panoel:
This "feels" wrong.  Should be a way to just get the currentNetwork.
So, leaving this as a draft until comments come in.


evan-gray:
Feels right now 😁

PR #90, opened 2023-05-12
dashboard: Added tx explorer links to Accountant component

kev1n-peters:


evan-gray (approved):
Tested with an Eth tx. Very nice. ➕ for code reuse. Might be good to have a helper methods do the `0x` enforcement at some point, maybe even get some types involved.

PR #93, opened 2023-06-28
Optimism update

panoel:
Do we want to remove the test for the old Optimism watcher?  I feel like eventually they will remove that old contract?


evan-gray (approved):
🙌

PR #94, opened 2023-06-29
watcher: handle karura code 6969

panoel:
This code handles the following from Karura:
```
      {
        id: '0',
        jsonrpc: '2.0',
        error: {
          code: 6969,
          message: 'Error: No response received from RPC endpoint in 60s'
        }
      },
```


evan-gray (approved):
ugh. okay.

PR #95, opened 2023-06-29
Terra2: bad block fix

panoel:
This fix is for when Terra finder says there is no transaction in a block but the lcd says that there is...


evan-gray (approved):
at quick glance it seems like this should catch the error we're seeing. also why is github diff so bad.

PR #96, opened 2023-06-30
Karura null blocks

panoel:
Sometimes Karura will reply with a null block on a perfectly valid block number.  We need to handle that.
Request:
```
    {
      jsonrpc: '2.0',
      id: '1',
      method: 'eth_getBlockByNumber',
      params: [ '0x461e9d', false ]
    }
```
Response:
```
 { id: '1', jsonrpc: '2.0', result: null } 
```

### DIFF
--- a/watcher/src/watchers/EVMWatcher.ts
+++ b/watcher/src/watchers/EVMWatcher.ts
@@ -28,10 +28,12 @@ export type ErrorBlock = {
 export class EVMWatcher extends Watcher {
   finalizedBlockTag: BlockTag;
   lastTimestamp: number;
+  latestFinalizedBlockNumber: number;
 
   constructor(chain: EVMChainName, finalizedBlockTag: BlockTag = 'latest') {
     super(chain);
     this.lastTimestamp = 0;
+    this.latestFinalizedBlockNumber = 0;
     this.finalizedBlockTag = finalizedBlockTag;
     if (chain === 'acala' || chain === 'karura') {
       this.maximumBatchSize = 50;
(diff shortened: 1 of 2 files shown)


watcher/src/watchers/EVMWatcher.ts
```diff
@@ -62,7 +62,17 @@ export class EVMWatcher extends Watcher {
         AXIOS_CONFIG_JSON
       )
     )?.data?.[0];
-    if (result && result.error && result.error.code === 6969) {
+    if (result && result.result === null) {
+      // Found null block
+      const latestBlockNumber = await this.getFinalizedBlockNumber();
+      if (typeof blockNumberOrTag === 'number' && blockNumberOrTag < latestBlockNumber - 1000) {
```
evan-gray, line 68: should the `typeof` part of this check be before calling `getFinalizedBlockNumber`? no point in calling if we won't use the result

watcher/src/watchers/EVMWatcher.ts
```diff
@@ -98,13 +108,17 @@ export class EVMWatcher extends Watcher {
       });
     }
     const results = (await axios.post(rpc, reqs, AXIOS_CONFIG_JSON))?.data;
+    const latestBlockNumber = await this.getFinalizedBlockNumber();
```
evan-gray, line 111: it's kinda awkward that this would always get called twice now in the watcher loop. should we cache `latestBlockNumber`the way you did `lastTimestamp`? that would avoid that extra call in `getBlock` as well.

evan-gray (changes requested):
one small optimization request. also not immediately sure why tests failed.

watcher/src/watchers/EVMWatcher.ts
panoel: Fixed by making latestFinalizedBlockNumber a member variable.

watcher/src/watchers/EVMWatcher.ts
panoel: Fixed by making latestFinalizedBlockNumber a member variable.

evan-gray (approved):

PR #98, opened 2023-07-06
DRAFT: initial pretty-fly implementation

barnjamin:
While working with the fly I found myself wanting to just implement some little chunk of logic given a certain message type.

This implementation provides a `Fly` interface that can be used to handle messages of any type without the need to spin up goroutines and handle context and logging logic.

This may be useful for reimplementing some of the other tools in the `fly` directory


prettyfly/cmd/fly/config.go
```diff
@@ -0,0 +1,23 @@
...
+	P2pNetworkID       string
+	P2pPort            uint
+	P2pBootstrap       string
+	NodeKeyPath        string
+	LogLevel           string
+	SignedVAATopicName string
+	ChannelBuffer      int
+}
+
+func DefaultConfig() Config {
```
barnjamin, line 13: should have one to read from a file?

PR #120, opened 2023-08-30
cloud_function: add alarmMissingVAAs

panoel:
Not sure how environment variables work in cloud functions.  If there needs to be code changes, I haven't done them.

Also, calling myself out here:  `CHAIN_INFO_MAP` is in 2 different consts files.  But they are slightly different.  Is there a better way?


cloud_functions/src/alarmMissingVaas.ts
```diff
@@ -0,0 +1,51 @@
...
+  } catch (e) {
+    console.log('could not get missing VAAs');
+    res.sendStatus(500);
+  }
+  return;
+}
+
+function formatMessage(msg: ObservedMessage): string {
+  let displayInfo: DisplayInfo = {
+    txHash: msg.txHash,
```
bruce-riley, line 38: Could you add the chainName here to make it easier for humans to understand?

panoel: That and more...

common/src/consts.ts
```diff
@@ -110,3 +133,285 @@ export const CIRCLE_DOMAIN_TO_CHAIN_ID: { [key: number]: ChainId } = {
...
+    explorerStem: `https://basescan.org`,
+  },
+  32: {
+    name: 'sei',
+    evm: false,
+    chainId: CHAIN_ID_SEI,
+    endpointUrl: '',
+    explorerStem: '',
+  },
+  33: {
```
evan-gray, line 327: This ID doesn't seem right. did you mean 3104

dashboard/src/utils/consts.ts
```diff
@@ -276,6 +277,14 @@ export const CHAIN_INFO_MAP: { [key: string]: CHAIN_INFO } = {
     explorerStem: '',
     icon: seiIcon,
   },
+  33: {
```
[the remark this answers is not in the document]
evan-gray: actually, should this whole thing just be imported from the consts `CHAIN_INFO_MAP`? 🤔

evan-gray (changes requested):
good start, couple small things in the dashboard 🙂

dashboard/src/utils/consts.ts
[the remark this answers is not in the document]
panoel: Put most of this in common/consts.ts and created new structure for icon images.

PR #121, opened 2023-08-31
watcher: wormchain support

evan-gray:


watcher/src/watchers/CosmwasmWatcher.ts
```diff
@@ -14,7 +14,7 @@ export class CosmwasmWatcher extends Watcher {
   rpc: string | undefined;
   latestBlockHeight: number;
 
-  constructor(chain: CosmWasmChainName) {
+  constructor(chain: CosmWasmChainName | 'wormchain') {
```
panoel, line 17: Should wormchain be a CosmWasmChainName?

panoel (dismissed):
So, it looks like standard cosmwasm watcher pretty much just worked?

watcher/src/watchers/CosmwasmWatcher.ts
evan-gray: It should, but it doesn't appear to be in the latest version of the sdk. idk if I can make an issue straight from here to the other repo

evan-gray:
> So, it looks like standard cosmwasm watcher pretty much just worked?

Almost, except that the responses weren't quite the same so I had to copy a bunch. We should see if we can go back and clean up the accessor difference later.

panoel (approved):

PR #129, opened 2023-09-12
Adding step to lookup addresses from lookup table

barnjamin:


watcher/scripts/checkSolana.ts
```diff
@@ -0,0 +1,14 @@
...
+dotenv.config();
+
+import { SolanaWatcher } from '../src/watchers/SolanaWatcher';
+import { CONFIG } from '@wormhole-foundation/connect-sdk';
+
+// Temporary script to test SolanaWatcher for lookup addresses
+(async () => {
+  const conf = CONFIG.Testnet.chains.Solana!;
+  const sw = new SolanaWatcher(conf.rpc, conf.contracts.coreBridge);
+  const msgs = await sw.getMessagesForBlocks(242637716, 242637736);
```
barnjamin, line 11: the slot with the transaction in question is `242637719`

watcher/src/watchers/SolanaWatcher.ts
```diff
@@ -31,71 +32,76 @@ export class SolanaWatcher extends Watcher {
...
   maximumBatchSize = 100_000;
 
-  constructor() {
+  connection: Connection | undefined;
+
+  constructor(rpc?: string, contract?: string) {
     super('solana');
-    this.rpc = RPCS_BY_CHAIN.solana!;
+    this.rpc = rpc ?? RPCS_BY_CHAIN.solana!;
+    this.programId = contract ?? WORMHOLE_PROGRAM_ID;
```
barnjamin, line 40: this is not exactly related to this change, so should take out (?) but makes it easier to test

watcher/src/watchers/SolanaWatcher.ts
```diff
@@ -31,71 +32,76 @@ export class SolanaWatcher extends Watcher {
...
   }
 
-  async getFinalizedBlockNumber(): Promise<number> {
-    const connection = new Connection(this.rpc, COMMITMENT);
-    return connection.getSlot();
+  getConnnection(): Connection {
```
barnjamin, line 43: i don't like creating a new conn for each fn, rather create once and cache

watcher/src/watchers/SolanaWatcher.ts
```diff
@@ -31,71 +32,76 @@ export class SolanaWatcher extends Watcher {
...
-    // in the rare case of maximumBatchSize skipped blocks in a row,
-    // you might hit this error due to the recursion below
-    if (fromSlot > toSlot) throw new Error('solana: invalid block range');
-    this.logger.info(`fetching info for blocks ${fromSlot} to ${toSlot}`);
-    const vaasByBlock: VaasByBlock = {};
+  async getFinalizedBlockNumber(): Promise<number> {
+    return this.getConnnection().getSlot();
+  }
 
+  private async findNextValidBlock(slot: number, next: number): Promise<VersionedBlockResponse> {
```
barnjamin, line 52: move recursive stepping thru slots to stand alone fn

watcher/src/watchers/SolanaWatcher.ts
```diff
@@ -129,37 +137,65 @@ export class SolanaWatcher extends Watcher {
...
+            // Important to return the addresses in the order they're specified in the
+            // address table lookup object. Note writable comes first, then readable.
+            return atl.writableIndexes
+              .concat(atl.readonlyIndexes)
+              .map((i) => lookupTableAccount.state.addresses[i]);
+          });
+
+          const lookups = await Promise.all(lookupPromises);
+          accountKeys = accountKeys.concat(...lookups);
+        }
```
barnjamin, line 164: Meat of the change, check the tx message for lookups, if theyre specified, get the lookup table and append the addresses (IN ORDER) to the `accountKeys`

Untested: multiple lookups, unsure if the ordering here will break it e.g. if they both specify writable and those writable indices must still appear first

watcher/src/watchers/SolanaWatcher.ts
```diff
@@ -129,37 +137,65 @@ export class SolanaWatcher extends Watcher {
...
+
         const whInstructions = innerInstructions
           .concat(instructions)
           .filter((i) => i.programIdIndex === programIdIndex);
+
         for (const instruction of whInstructions) {
           // skip if not postMessage instruction
           const instructionId = instruction.data;
-          if (instructionId[0] !== 0x01) continue;
+          if (instructionId[0] !== 0x08) continue;
```
barnjamin, line 180: https://explorer.solana.com/tx/4hcuEcdk2J2dcgjkun9ytEmcmYAp7ts9fdbJoaHmYjRJdXu1YzhfBeYtmtQ2UE6NFQumqVkQJNW1qFLMqTT374kN?cluster=devnet

step 2.8, passes 0x08 as first byte?

watcher/src/watchers/SolanaWatcher.ts
evan-gray: seems like this should be a default on the parameter

watcher/src/watchers/SolanaWatcher.ts
```diff
@@ -31,71 +32,89 @@ export class SolanaWatcher extends Watcher {
...
+    const vaasByBlock: VaasByBlock = {};
+
+    // identify block range by fetching signatures of the first and last transactions
+    // getSignaturesForAddress walks backwards so fromSignature occurs after toSignature
+
+    // start by finding a valid range of blocks so we can use their
+    // signatures in the `getSignaturesForAddress` search
+    // look for the (last block + 1) and (first block - 1) since the signature parameters in the search later
+    // are _exclusive_ so we have to get the signatures immediate preceeding or following the ones we're interested in
+    const retries = 5;
```
barnjamin, line 96: what should this be? is 5 ok? should we set it on init of watcher?

PR #132, opened 2023-09-19
fix: change url for solana tx to refer to storage account instead of tx

barnjamin:
image

links to https://solscan.io/tx/AdAZCfrAGFdVHfsNsPm7VRXwVmiuiD2nQg5YMD8QjdW1 but the "txid" we use is actually the storage account


evan-gray (approved):
approving since I _think_ this is true for all uses of this (though ew)

PR #135, opened 2023-09-20
cloud_functions: don't alarm governed VAAs

panoel:
Grab the governed VAAs from the firestore governorStatus table.  Add filtering so that these VAAs don't get alarmed in slack.


evan-gray (changes requested):
should be quicker to simply grab all the documents in the collection.

https://firebase.google.com/docs/firestore/query-data/get-data#node.js_6

barnjamin (approved):

barnjamin (approved):

cloud_functions/src/alarmMissingVaas.ts
```diff
@@ -64,6 +103,48 @@ export async function alarmMissingVaas(req: any, res: any) {
...
+// This function gets all the enqueued VAAs from he governorStatus collection.
+async function getGovernedVaas(): Promise<GovernedVAAMap> {
+  const vaas: GovernedVAAMap = new Map<string, GovernedVAA>();
+  // Walk all the guardians and retrieve the enqueued VAAs
+  const firestore = new Firestore();
+  const collection = firestore.collection(
+    assertEnvironmentVariable('FIRESTORE_GOVERNOR_STATUS_COLLECTION')
+  );
+  const snapshot = await collection.get();
+  snapshot.forEach(async (doc) => {
```
evan-gray, line 115: I think you can't use an async method in a forEach. I don't think you need the async since you don't have any awaits. I've found a better habit is using 
```
for (const doc of snapshot) {
  // do stuff
  // you can even await stuff in here
}
```

evan-gray (changes requested):

cloud_functions/src/alarmMissingVaas.ts
panoel: I stole this code from a different cloud_function.  We should probably change that one, too.
Fixed.

evan-gray (approved):

PR #136, opened 2023-09-22
cloud_functions: latestTokenData

panoel:


cloud_functions/src/getLatestTokenData.ts
```diff
@@ -0,0 +1,87 @@
...
+const REFRESH_TIME_INTERVAL =
+  Number(process.env.CLOUD_FUNCTIONS_REFRESH_TIME_INTERVAL) || 1000 * 60 * 5;
+
+export async function getLatestTokenData(req: any, res: any) {
+  res.set('Access-Control-Allow-Origin', '*');
+  if (req.method === 'OPTIONS') {
+    // Send response to OPTIONS requests
+    res.set('Access-Control-Allow-Methods', 'GET');
+    res.set('Access-Control-Allow-Headers', 'Content-Type');
+    res.set('Access-Control-Max-Age', '3600');
```
barnjamin, line 57: should we set this to a full day? not sure anything is paying attention to it but seems the max allowed is 86400

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age#directives

cloud_functions/src/getLatestTokenData.ts
```diff
@@ -0,0 +1,87 @@
...
+      console.log('After getLatestTokenData =', timeDiff);
+      cache['values'] = values;
+      cache['lastUpdated'] = Date.now();
+    } else {
+      console.log(`cache is still valid, not refreshing ${new Date()}`);
+      values = cache['values'];
+    }
+    res.status(200).send(JSON.stringify(values));
+  } catch (e) {
+    res.sendStatus(500);
```
barnjamin, line 85: log the error?

cloud_functions/src/getLatestTokenData.ts
panoel: This is the value set for most of the other cloud_functions.  Just being consistent. 
@evan-gray , what say you?

cloud_functions/src/getLatestTokenData.ts
panoel: Added logging.

evan-gray (dismissed):

barnjamin (approved):

PR #140, opened 2023-10-05
Near watcher to use backfill

panoel:
This PR is using a different approach to get the Near watcher to catch up.

### DIFF
--- a/watcher/scripts/backfillNear.ts
+++ b/watcher/scripts/backfillNear.ts
@@ -8,7 +8,12 @@ import {
 import { BlockResult, Provider } from 'near-api-js/lib/providers/provider';
 import ora from 'ora';
 import { initDb, makeBlockKey } from '../src/databases/utils';
-import { getNearProvider, getTransactionsByAccountId, NEAR_ARCHIVE_RPC } from '../src/utils/near';
+import {
+  getNearProvider,
+  getTimestampByBlock,
+  getTransactionsByAccountId,
+  NEAR_ARCHIVE_RPC,
+} from '../src/utils/near';
 import { getMessagesFromBlockResults } from '../src/watchers/NearWatcher';
 import { Transaction } from '../src/types/near';
 import { VaasByBlock } from '../src/databases/types';
@@ -22,11 +27,6 @@ import { VaasByBlock } from '../src/databases/types';
 // Otherwise, the script will backfill the local JSON database.
 
 const BATCH_SIZE = 100;
-
-async function getTimestampByBlock(provider: Provider, blockHeight: number): Promise<number> {
-  const block: BlockResult = await provider.block({ blockId: blockHeight });
-  return block.header.timestamp;
-}
 
 (async () => {
   const db = initDb(false); // Don't start watching
@@ -58,16 +58,15 @@ async function getTimestampByBlock(provider: Provider, blockHeight: number): Pro
     log.text = `Fetching blocks... ${i + 1}/${blockHashes.length}`;
     let success: boolean = false;
     while (!success) {
-      success = true;
       try {
         const block = await provider.block({ blockId: blockHashes[i] });
         if (block.header.height > fromBlock && block.header.height <= toBlock.header.height) {
           blocks.push(block);
         }
+        success = true;
       } catch (e) {
-        console.log('Error fetching block', e);
+        console.error('Error fetching block', e);
         await sleep(5000);
-        success = false;
       }
     }
   }
--- a/watcher/scripts/backfillNear.ts
+++ b/watcher/scripts/backfillNear.ts
@@ -1,12 +1,17 @@
 import * as dotenv from 'dotenv';
 dotenv.config();
 import { ChainName, CONTRACTS } from '@certusone/wormhole-sdk/lib/cjs/utils/consts';
-import { INITIAL_DEPLOYMENT_BLOCK_BY_CHAIN } from '@wormhole-foundation/wormhole-monitor-common';
-import { BlockResult } from 'near-api-js/lib/providers/provider';
+import {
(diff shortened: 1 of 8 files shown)


watcher/scripts/backfillNear.ts
```diff
@@ -44,14 +56,32 @@ const BATCH_SIZE = 100;
...
-    if (block.header.height > fromBlock && block.header.height <= toBlock.header.height) {
-      blocks.push(block);
+    let success: boolean = false;
+    while (!success) {
+      success = true;
+      try {
+        const block = await provider.block({ blockId: blockHashes[i] });
+        if (block.header.height > fromBlock && block.header.height <= toBlock.header.height) {
+          blocks.push(block);
+        }
```
barnjamin, line 66: can just put success=true here and not reassign in err condition

watcher/scripts/backfillNear.ts
```diff
@@ -44,14 +56,32 @@ const BATCH_SIZE = 100;
...
+      } catch (e) {
+        console.log('Error fetching block', e);
```
barnjamin, line 68: make this a console.error?

watcher/src/utils/near.ts
```diff
@@ -22,35 +23,75 @@ export const getNearProvider = async (rpc: string): Promise<Provider> => {
...
+    //     params
+    //   )}}]`
+    // );
+    const res = (
+      (
+        await axios.get(
+          `${NEAR_EXPLORER_TRANSACTION_URL}?batch=1&input={"0":${JSON.stringify(params)}}`,
+          AXIOS_CONFIG_JSON
+        )
+      ).data as GetTransactionsByAccountIdResponse
```
barnjamin, line 65: this is in a while loop but i dont immediately see how this request changes for each iteration?

barnjamin: jk i see params.cursor below

watcher/src/watchers/NearArchiveWatcher.ts
```diff
@@ -0,0 +1,146 @@
...
+    const block = await provider.block({ finality: 'final' });
+    return block.header.height;
+  }
+
+  async getMessagesForBlocks(fromBlock: number, toBlock: number): Promise<VaasByBlock> {
+    // assume toBlock was retrieved from getFinalizedBlockNumber and is finalized
+    this.logger.info(`fetching info for blocks ${fromBlock} to ${toBlock}`);
+    const provider = await this.getProvider();
+    const fromBlockTimestamp: number = await getTimestampByBlock(provider, fromBlock);
+    const toBlockInfo: BlockResult = await provider.block({ blockId: toBlock });
```
barnjamin, line 39: i want to suggest changing the getTimestampByBlock to take a BlockResult since it does its own fetch of a block but not sure its worth it

watcher/src/watchers/NearArchiveWatcher.ts
```diff
@@ -0,0 +1,146 @@
...
+        success = true;
+        try {
+          const block = await provider.block({ blockId: blockHashes[i] });
+          if (block.header.height > fromBlock && block.header.height <= toBlockInfo.header.height) {
+            blocks.push(block);
+          }
+        } catch (e) {
+          console.log('Error fetching block', e);
+          await sleep(5000);
+          success = false;
```
barnjamin, line 63: same comments for this one as the above, console.error, change where success gets set

barnjamin (approved):

watcher/scripts/backfillNear.ts
panoel: fixed

watcher/scripts/backfillNear.ts
panoel: Fixed

watcher/src/watchers/NearArchiveWatcher.ts
panoel: Fixed.

watcher/src/utils/near.ts
```diff
@@ -22,6 +24,32 @@ export const getNearProvider = async (rpc: string): Promise<Provider> => {
...
+  let block: BlockResult = {} as BlockResult;
+  while (!success) {
+    try {
+      block = await provider.block({ blockId: blockHeight });
+      success = true;
+    } catch (e) {
+      console.error('Error fetching block', e);
+      await sleep(5000);
+    }
+  }
```
barnjamin, line 49: should we have a condition for max retries or max time?

watcher/src/watchers/NearArchiveWatcher.ts
```diff
@@ -0,0 +1,146 @@
...
+  provider: Provider | null = null;
+
+  constructor() {
+    super('near');
+  }
+
+  async getFinalizedBlockNumber(): Promise<number> {
+    this.logger.info(`fetching final block for ${this.chain}`);
+    const provider = await this.getProvider();
+    const block = await provider.block({ finality: 'final' });
```
barnjamin, line 31: use the new method?

watcher/src/utils/near.ts
panoel: I thought about that.  But there is really nothing else we can do but wait.  The error message every 5 seconds will show up in my logging monitoring.  So, it becomes more apparent.  In other places, we have exponential backoff.  But what happens is that it backs off so much that the errors get lost.

watcher/src/watchers/NearArchiveWatcher.ts
panoel: The *finality* piece is different.  That is why I called the new function ...ByBlockId().  I feel like it is okay in this instance because it is not running in a loop where it could lose all those previous lookups.  If you feel we *really* should add this to the other function, I'll add it.

barnjamin (approved):

PR #146, opened 2023-10-23
watcher: add checks for base64 in cosmwasm

panoel:
Some cosmwasm chains are using base64 encoding and some are not.  This PR allows for both.

### DIFF
--- a/watcher/src/utils/isBase64Encoded.ts
+++ b/watcher/src/utils/isBase64Encoded.ts
@@ -1,0 +1,13 @@
+// This function uses a regex string to check if the input could
+// possibly be base64 encoded.
+//
+// WARNING:  There are clear text strings that are NOT base64 encoded
+//           that will pass this check.
+export function isBase64Encoded(input: string): boolean {
+  const b64Regex = new RegExp('^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$');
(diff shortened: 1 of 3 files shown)


watcher/src/utils/isBase64Encoded.ts
```diff
@@ -0,0 +1,13 @@
...
+// WARNING:  There are clear text strings that are NOT base64 encoded
+//           that will pass this check.
+export function isBase64Encoded(input: string): boolean {
+  const b64Regex = new RegExp('^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$');
+  const match = b64Regex.exec(input);
+  if (match) {
+    return true;
+  }
+  return false;
+}
```
barnjamin, line 13: How about just a `try/catch` around b64decode for msg/value? if the try/catch fails, return the values you're passed

barnjamin: eg:

```ts
export function maybeB64Decode(k,v):[string,string] {
...
}
```

then to use it
```ts
const [key, value] = maybeB64Decode(attrs[k].key, attrs[k].value)
```

panoel: Added a helper function

barnjamin (approved):

barnjamin (approved):

PR #161, opened 2023-12-08
watcher: update guardian rpc hosts

panoel:


bruce-riley (approved):

barnjamin (approved):

watcher/src/consts.ts
```diff
@@ -84,7 +84,7 @@ export const AXIOS_CONFIG_JSON: AxiosRequestConfig = {
 };
 
 export const GUARDIAN_RPC_HOSTS = [
-  'https://wormhole-v2-mainnet-api.certus.one',
+  'https://api.wormholescan.io',
```
kev1n-peters, line 87: Does this API provide the methods that we would be calling on the guardian?

barnjamin: Looks like we only use the one signed_vaa endpoint and that is supported 👍 

https://github.com/wormhole-foundation/wormhole-dashboard/blob/main/watcher/scripts/fetchMissingVAAs.ts#L40-L44

https://github.com/wormhole-foundation/wormhole-dashboard/blob/main/watcher/src/utils/getSignedVAA.ts#L9-L14

PR #163, opened 2023-12-12
fly: put tunables into .env file

panoel:
This PR removes some hard coded tunables and expects them to be in a .env file.

### DIFF
--- a/fly/cmd/fly/main.go
+++ b/fly/cmd/fly/main.go
@@ -3,7 +3,6 @@ package main
 import (
 	"context"
 	"encoding/hex"
-	"flag"
 	"fmt"
 	"os"
 	"strconv"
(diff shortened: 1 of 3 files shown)


fly/cmd/fly/main.go
```diff
@@ -107,28 +110,38 @@ func incrementPythNetMsgCount(ctx context.Context, client *firestore.Client, cou
 	}
 }
 
+func loadEnvVars() {
+	err := godotenv.Load() // By default loads .env
+	if err != nil {
+		log.Fatal("Error loading .env file")
+	}
+	p2pNetworkID = os.Getenv("P2P_NETWORK_ID")
```
bruce-riley, line 118: Should you err out if `P2P_NETWORK_ID` or `P2P_BOOTSTRAP` are empty?

Same for all the other ones, I guess. Maybe create a function that returns the string but does a `log.Fatal` if it's not set. (Of course maybe some of these can be empty?)

kev1n-peters (approved):

bruce-riley (approved):

fly/cmd/fly/main.go
panoel: Good call.

bruce-riley (approved):

PR #165, opened 2023-12-17
cloud-functions: Increased coingecko delay and chunk size

kev1n-peters:


kev1n-peters:
As the number of tokens wormhole tracks has increased, CoinGecko was starting to rate limit us when fetching prices.

panoel (approved):

PR #167, opened 2023-12-19
Add testnet monitor with fly, watcher, dashboard

panoel:


watcher/src/watchers/MoonbeamWatcher.ts
```diff
@@ -1,27 +1,30 @@
...
   }
   async getFinalizedBlockNumber(): Promise<number> {
     const latestBlock = await super.getFinalizedBlockNumber();
     let isBlockFinalized = false;
+    if (!RPCS_BY_CHAIN[this.network].moonbeam) {
+      throw new Error('Moonbeam RPC is not defined!');
+    }
+    const rpc: string = (
+      RPCS_BY_CHAIN[this.network].moonbeam ? RPCS_BY_CHAIN[this.network].moonbeam : ''
+    ) as string;
```
evan-gray, line 18: If you set the variable first, the type refinement will work :+1:

```suggestion
    const rpc = RPCS_BY_CHAIN[this.network].moonbeam
    if (!rpc) {
      throw new Error('Moonbeam RPC is not defined!');
    }
```

evan-gray (dismissed):
seems like a reasonable refactor, just one small nit.

PR #180, opened 2024-01-10
dashboard: handle empty array

panoel:


dashboard/src/components/Monitor.tsx
```diff
@@ -173,7 +173,10 @@ function DetailBlocks({ chain }: { chain: string }) {
       cancelled = true;
     };
   }, [chain, fromId, currentNetwork]);
-  const rawMessages = messagesWrapper.data;
+  let rawMessages = messagesWrapper.data;
+  if (rawMessages && rawMessages.length === 0) {
+    rawMessages = null;
+  }
```
evan-gray, line 179: This works. I might have done
```suggestion
  const rawMessages = messagesWrapper.data?.length === 0 ? null : messagesWrapper.data;
```
because :shrug:

evan-gray (approved):

dashboard/src/components/Monitor.tsx
panoel: Understood.

PR #184, opened 2024-01-12
dashboard: change explorer rpc

panoel:
I did actually run the dashboard to make sure the explorer link showed up correctly and went to the correct location.


evan-gray (approved):

evan-gray:
tested locally, there's an issue with the default moonbeam rpc

PR #185, opened 2024-01-16
fly: Updated p2p package for cutover support

kev1n-peters:


evan-gray (approved):

SEJeff (approved):
Confirmed the bootstrap url.

It might make sense to make a package that has the bootstrap url and just share it as a variable between your various cmd packages fwiw.

PR #210, opened 2024-02-21
watcher: fix aptos emitter format

panoel:


bingyuyap (dismissed):
Approved with small comment

watcher/src/watchers/AptosWatcher.ts
```diff
@@ -61,7 +61,8 @@ export class AptosWatcher extends Watcher {
         ]);
         const timestamp = new Date(Number(block.block_timestamp) / 1000).toISOString();
         const blockKey = [block.block_height, timestamp, sequence_number].join('/'); // use custom block key for now so we can include sequence number
-        const emitter = data.sender.padStart(64, '0');
+        // Data.sender is a decimal as a string.  It needs to be hexadecimal.
+        const emitter = Number(data.sender).toString(16).padStart(64, '0');
```
bingyuyap, line 65: is `data.sender` a very large number? Might be safer to use `BigInt(data.sender)` instead of `Number(data.sender)` especially if `data.sender` exceeds `Number.MAX_SAFE_INTEGER` 

```ts
const emitter = BigInt(data.sender).toString(16).padStart(64, '0');
```

panoel: Typically there won't be that many emitters.  But it doesn't hurt to be safe.

PR #211, opened 2024-02-27
historical_uptime: push precomputed chain height differences

bingyuyap:
Precompute chain height differences before pushing to prometheus. This makes it easier to group and reduce the metrics on Grafana side.

### DIFF
--- a/fly/cmd/historical_uptime/main.go
+++ b/fly/cmd/historical_uptime/main.go
@@ -59,6 +59,14 @@ var (
 		[]string{"guardian", "chain"},
 	)
 
+	guardianChainHeightDifferences = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "guardian_chain_height_differences",
+			Help: "Current height difference of each guardian from max height on each chain over time",
+		},
+		[]string{"guardian", "chain"},
+	)
+
 	guardianHeartbeats = promauto.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "guardian_heartbeats",
@@ -77,6 +85,11 @@ var (
 )
 
 const PYTHNET_CHAIN_ID = int(vaa.ChainIDPythNet)
+
+var (
+	// guardianChainHeights indexes current chain height by chain id and guardian name
+	guardianChainHeights = make(common.GuardianChainHeights)
+)
 
 func loadEnvVars() {
 	err := godotenv.Load() // By default loads .env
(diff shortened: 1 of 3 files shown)


fly/pkg/historical_uptime/helpers.go
```diff
@@ -69,3 +69,43 @@ func UpdateMetrics(guardianMissedObservations *prometheus.CounterVec, guardianMi
 		}
 	}
 }
+
+func computeMaxChainHeights(guardianChainHeights map[uint32]map[string]uint64) map[uint32]uint64 {
```
panoel, line 73: Could we typedef these maps, please? e.g.
```
type GuardianHeight map[string]uint64 // map of guardian addr to chain height
type GuardianChainHeights map[uint32]GuardianHeight // map of chainIds to guardian heights
```

bingyuyap: will do when im back, thanks!

fly/pkg/historical_uptime/helpers.go
```diff
@@ -69,3 +69,43 @@ func UpdateMetrics(guardianMissedObservations *prometheus.CounterVec, guardianMi
...
+
+func computeGuardianChainHeightDifferences(guardianChainHeights map[uint32]map[string]uint64, maxChainHeights map[uint32]uint64) map[uint32]map[string]uint64 {
+	heightDifferences := make(map[uint32]map[string]uint64)
+
+	for chainId, guardianHeights := range guardianChainHeights {
+		for guardian, height := range guardianHeights {
+			if heightDifferences[chainId] == nil {
+				heightDifferences[chainId] = make(map[string]uint64)
+			}
+			heightDifferences[chainId][guardian] = maxChainHeights[chainId] - height
```
panoel, line 99: Is maxChainHeights[chainId] always guaranteed to be >= height so that you don't have to deal with underflow?

fly/cmd/historical_uptime/main.go
```diff
@@ -324,6 +350,11 @@ func main() {
 				}
 
 				for _, network := range hb.Networks {
+					if guardianChainHeights[network.Id] == nil {
```
panoel, line 353: You may get some weird chain IDs.  Do you want to reject network.Id, if it isn't a chain we care about?

panoel: Oh, wait, these chains are contained in the guardian HB, right?  In that case, you should be good here.

panoel (dismissed):

fly/cmd/historical_uptime/main.go
bingyuyap: I left it in because we can filter on Grafana side, which is easier to config.

fly/pkg/historical_uptime/helpers.go
bingyuyap: yes, since in `computeMaxChainHeights` (right before we invoke `computeGuardianChainHeightDifferences`) we compute the max height in every chain

```go
if highest < guardianHeight {
  highest = guardianHeight
}
```

fly/pkg/historical_uptime/helpers.go
bingyuyap: added typedef in `common/types.go`

panoel (approved):

PR #213, opened 2024-03-06
historical_uptime: dynamic batch update and delete messages entirely …

bingyuyap:
…to reduce memory usage


bingyuyap:
Closing since we are not using BadgerDB anymore.

PR #214, opened 2024-03-06
watcher: add NTT watcher

panoel:


watcher/src/watchers/NTTWatcher.ts
```diff
@@ -0,0 +1,747 @@
...
+          }
+        } else if (log.topics[0] === InboundTransferQueuedTopic) {
+          this.logger.debug('***********InboundTransferQueuedTopic***************');
+          let digest: string = log.data;
+          if (digest.startsWith('0x')) {
+            digest = digest.slice(2);
+          }
+          this.logger.debug('digest:', digest);
+          // Check if we have a lifecycle for this digest
+          const lc = this.lifecycleMap.get(digest);
```
bingyuyap, line 468: I wonder if we can just update the table by digest, instead of getting it from the map. What if we restart the service halfway and the data in the in memory map is lost? this `lc` would be `null`?

watcher/src/watchers/NTTWatcher.ts
```diff
@@ -0,0 +1,747 @@
...
+  }
+  // Expected input format is:2024-03-01T02:30:45.000Z
+  // Convert the 'T' to a space
+  let parts = timestamp.split('T');
+  // Remove the trailing 'Z'
+  parts[1] = parts[1].slice(0, -1);
+  return parts.join(' ');
+}
+
+function mergeLifecycles(lc: LifeCycle, existing: LifeCycle): LifeCycle {
```
bingyuyap, line 687: I think we can delete this function now that it is not used.

watcher/src/watchers/NTTWatcher.ts
```diff
@@ -0,0 +1,747 @@
...
+        .update({
+          outbound_transfer_releasable_time: formatIntoTimestamp(lc.outboundTransferReleasableTime),
+        });
+    } else {
+      logger.error(`saveToPG: Unknown initiating event: ${initiatingEvent} and lifeCycle: ${lc}`);
+    }
+  });
+}
+
+function formatIntoTimestamp(timestamp: string): string {
```
bingyuyap, line 675: there's a `formatIntoTimestamp` in `watcher/utils/timestamp.ts` as well. When we have the time we can reduce the code duplication. We can do it in another PR if we want.

bingyuyap (dismissed):
Looks good! Just 2 very small comments. I have looked through the code a couple of times (not to mention the other times during development) and don't think there's any issues. Hopefully didn't miss anything important. 

Thank you for the hard work!

watcher/src/watchers/NTTWatcher.ts
panoel: Actually, lifecycleMap was just a way of prototyping this code before adding the postgres db.  lifecycleMap should be completely removed, now.

watcher/src/watchers/NTTWatcher.ts
panoel: Removed.

PR #228, opened 2024-03-24
historical_uptime: update logic to use bigtable

bingyuyap:
This PR includes:
1. bigtable functions to save, update, delete messages, observations, messageIndexes
2. updating main logic to use bigtable instead of badgerdb
3. test cases
4. delete badgerdb code

Most of the code is related to bigtable mutation and parsing. A fair bit of testing (517 lines). And deleted the BadgerDB code (529 lines). 

Previous PR is closed since most of the code are different, reviewing that would be unnecessary.

TODO:
1. optimize the bigtable usage, right now it's pinging the bigtable every observation. We might want to explore aggregating this and bulk update the bigtable at a regular time interval.


fly/cmd/historical_uptime/main.go
```diff
@@ -189,20 +206,32 @@ func initObservationScraper(db *db.Database, logger *zap.Logger, errC chan error
...
+				messages, err := db.GetUnprocessedMessagesBeforeCutOffTime(ctx, time.Now().Add(-common.ExpiryDuration))
 				if err != nil {
 					logger.Error("QueryMessagesByIndex error", zap.Error(err))
 					continue
 				}
 
+				for _, message := range messages {
+					observations, err := db.GetObservationsByMessageID(ctx, string(message.MessageID))
+					if err != nil {
+						logger.Error("GetObservationsByMessageID error", zap.Error(err))
```
panoel, line 220: Do you want to log the MessageID in case it is useful?

bingyuyap: You are right, I have added the messageID in the log. Thanks!

fly/pkg/bigtable/message.go
```diff
@@ -0,0 +1,243 @@
...
+// Here we do not need to filter by `metricsChecked“ field since we assume that the messages in messageIndex table have the `metricsChecked` field set to false.
+func (db *BigtableDB) GetUnprocessedMessagesBeforeCutOffTime(ctx context.Context, cutOffTime time.Time) ([]*types.Message, error) {
+	// Step 1: Get all message indexes
+	messageIndexes, err := db.GetAllMessageIndexes(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get message indexes: %v", err)
+	}
+
+	// Step 2: Use the row keys from messageIndex to get the messages
+	var rowKeys []string
```
bruce-riley, line 153: Maybe do `make([]string, len(messageIndexes)` then just do `rowKeys[idx] = string(index)`? Would require fewer mallocs.

fly/pkg/bigtable/message.go
```diff
@@ -0,0 +1,243 @@
...
+	var checkedMessagesId []types.MessageID
+	for _, message := range messages {
+		if message.LastObservedAt.Before(cutOffTime) {
+			filteredMessages = append(filteredMessages, message)
+			checkedMessagesId = append(checkedMessagesId, message.MessageID)
+			message.MetricsChecked = true
+		}
+	}
+
+	// Step 4: Mark the messages as processed
```
bruce-riley, line 174: Should this be moved below Step 5, so you don't mark them processed if the delete fails?

fly/pkg/bigtable/message.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// This works on the assumption that there is only one version of the message.
+// If there are multiple versions of the message, this will return which version is read last.
+func (db *BigtableDB) bigtableRowToMessage(row bigtable.Row) (*types.Message, error) {
+	var message types.Message
+	message.MessageID = types.MessageID(row.Key())
+
+	for _, column := range row["messageData"] {
+		switch column.Column {
```
bruce-riley, line 196: Not sure how many columns there are, or how they are ordered, but would it be worth breaking out of the loop once you've processed the ones you care about?

fly/pkg/bigtable/observation.go
```diff
@@ -0,0 +1,167 @@
...
+	var observation types.Observation
+	messageId, guardianAddr, err := db.parseObservationRowKey(row.Key())
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse observation row key: %v", err)
+	}
+	observation.MessageID = messageId
+	observation.GuardianAddr = guardianAddr
+
+	for _, column := range row["observationData"] {
+		switch column.Column {
```
bruce-riley, line 98: Similar comment about possibly breaking out when you have everything you care about.

bruce-riley (approved):
Looks good. Just a few minor comments. You can feel free to mark them resolved or do them in a later PR if you like.

fly/pkg/bigtable/message_index.go
```diff
@@ -0,0 +1,113 @@
...
+
+	err := db.client.Open(tableName).Apply(ctx, rowKey, mut)
+	if err != nil {
+		return fmt.Errorf("failed to delete message index: %v", err)
+	}
+
+	return nil
+}
+
+// BulkDeleteMessageIndex deletes the message index for a batch of messageIDs. This is used when the messages are processed and
```
panoel, line 52: nit: wrong name

panoel (dismissed):

fly/pkg/bigtable/message.go
```diff
@@ -0,0 +1,243 @@
...
+func NewBigtableDB(ctx context.Context, projectID, instanceID, credentialsFile, emulatorHost string, useBigtableEmulator bool) (*BigtableDB, error) {
+	var client *bigtable.Client
+	var err error
+
+	if !useBigtableEmulator && credentialsFile != "" {
+		client, err = bigtable.NewClient(ctx, projectID, instanceID, option.WithCredentialsFile(credentialsFile))
+	} else if useBigtableEmulator && emulatorHost != "" {
+		client, err = bigtable.NewClient(ctx, projectID, instanceID, option.WithoutAuthentication(), option.WithEndpoint(emulatorHost))
+	} else {
+		return nil, fmt.Errorf("invalid Bigtable configuration, if using emulator, set emulatorHost, else set credentialsFile")
```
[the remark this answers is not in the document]
bingyuyap: For this PR there's only 1 that requires this change! Will update. Thanks!

fly/pkg/bigtable/message.go
bingyuyap: Updated!

fly/pkg/bigtable/message.go
bingyuyap: Good call, I might want to implement a saga pattern to revert the deletion if save messages fail in near future too.

fly/pkg/bigtable/message.go
bingyuyap: Since we are persisting the columns with a timestamp of `0`, we would only have 1 column for each field. In this case we have 2 columns only. In this case the performance gain from breaking out of the loop early would be minimal. 

Will keep in mind that this is a point of optimization in future features. Thanks!

fly/pkg/bigtable/message_index.go
bingyuyap: Great catch. Thanks!

fly/pkg/bigtable/observation.go
bingyuyap: Observation, similarly uses timestamp `0` and only has 3 columns. So the performance gain would be minimal!

bruce-riley (approved):

PR #232, opened 2024-03-27
ntt_solana_watcher: add `is_relay`

bingyuyap:
NTT Solana Watcher updates:
- now watches both NTT Program and also NTT Quoter Program 
- fetches `outbox_item` account from `requestRelay` ix and checks if there is a relay requested 

To support this we also:
- add `ntt_quoter` IDL and types
- new table to store `outbox_item` <-> `digest` since it is required to know which digest corresponds to the outboxItem in `requestRelay` 
- solana `NttQuoter` class implementation 
- small test case to check parsing works 

Note: latest Solana contract is not deployed on `devnet`. This affects the `outbox_item` account index on the `requestRelay` ix by 1. Once it is deployed, we need to update this index in `parseRequestRelay`


watcher/src/NTTConsts.ts
```diff
@@ -161,6 +163,14 @@ export const NTT_CONTRACT: { [key in Environment]: { [key in ChainName]?: string
   ['devnet']: {},
 };
 
+export const NTT_QUOTER_CONTRACT: { [key in Environment]: { [key in ChainName]?: string } } = {
+  ['mainnet']: {},
+  ['testnet']: {
+    solana: 'NqTdGLLL6b6bFo7YESNEezocgF8onH5cst5EdH791en',
```
panoel, line 169: Could there ever be multiple quoter contracts, or just only one?

bingyuyap: Confirmed with Csongor that there's only one for all the NTT instances!

panoel (approved):

PR #243, opened 2024-04-01
common: move to sdk v2

panoel:


cloud_functions/src/processVaa.ts
```diff
@@ -69,7 +70,7 @@ export const processVaa: EventFunction = async (message: PubsubMessage, context:
       return;
     }
     const [chain, emitter] = rowKey.split('/');
-    const chainId = Number(chain);
+    const chainId = Number(chain) as ChainId;
```
barnjamin, line 73: `toChainId](https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/core/base/src/constants/chains.ts#L86) should parse and set the type

common/src/consts.ts
```diff
@@ -209,189 +178,189 @@ export const CHAIN_INFO_MAP: { [key in Environment]: { [key: string]: CHAIN_INFO
     1: {
       name: 'solana',
```
barnjamin, line 179: do these names need to change?

common/src/consts.ts
```diff
@@ -209,189 +178,189 @@ export const CHAIN_INFO_MAP: { [key in Environment]: { [key: string]: CHAIN_INFO
...
       evm: false,
-      chainId: CHAIN_ID_SOLANA,
+      chainId: toChainId('Solana'),
       endpointUrl: process.env.REACT_APP_SOLANA_RPC || 'https://api.mainnet-beta.solana.com',
```
barnjamin, line 182: can get this from rpc consts as well

watcher/src/watchers/InjectiveExplorerWatcher.ts
```diff
@@ -1,4 +1,4 @@
-import { CONTRACTS } from '@certusone/wormhole-sdk/lib/cjs/utils/consts';
+import { CONTRACTS, ChainName } from '@certusone/wormhole-sdk/lib/cjs/utils/consts';
```
bingyuyap, line 1: There should be a method `getContracts` from the `sdk-definitions` package that serves the same purpose. Since we are trying to move entire to `sdkv2` it might be worth updating this as well. You can check out [`getContracts`](https://github.com/wormhole-foundation/connect-sdk/blob/sdkv2/core/definitions/src/contracts.ts#L24) to see if this is what we want. 

```ts
import { getContracts } from '@wormhole-foundation/sdk-definitions';

// line 52
const coreAddress = getContracts(this.network, this.chain).coreBridge;
```

The thing about using this is that the `network` param in `getContracts`  uses `["Mainnet", "Testnet", "Devnet"]` but our repo uses `'mainnet' | 'testnet' | 'devnet'`. We can do this in another PR if you want.

bingyuyap (dismissed):
Everything looks good to me. Just one comment about using `getContracts` from `@wormhole-foundation/sdk-definitions` instead of `CONTRACTS` from the `certusone` sdk. 

Appreciate the work!

cloud_functions/src/processVaa.ts
evan-gray: can follow up when the rest of the methods are switched over

common/src/consts.ts
evan-gray: removed

common/src/consts.ts
evan-gray: removed

watcher/src/watchers/InjectiveExplorerWatcher.ts
evan-gray: let's tackle the watcher in a separate PR

evan-gray (approved):

evan-gray:
approving to get this change which replaces the deprecated sdk for the new one in `common` and `dashboard` only

evan-gray (changes requested):
ah, wait `toChain` isn't safe to use in all these places as it could throw

barnjamin (approved):

evan-gray (approved):
fixed `toChain` usage

PR #246, opened 2024-04-03
watcher: add normalized amount to NTT evm

panoel:
I moved common code around so there isn't so much copy pasta.  That's why it looks like a lot of code changes.

### DIFF
--- a/watcher/src/watchers/NTTWatcher.ts
+++ b/watcher/src/watchers/NTTWatcher.ts
@@ -346,14 +346,14 @@ export class NTTWatcher extends Watcher {
               );
               const sourceToken: string =
                 transceiverMessage.ntt_managerPayload.payload.sourceToken.toString('hex');
-              const decimals = transceiverMessage.ntt_managerPayload.payload.trimmedAmount.decimals;
-              const amount = transceiverMessage.ntt_managerPayload.payload.trimmedAmount.amount;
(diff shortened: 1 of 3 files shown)


watcher/src/watchers/NTTWatcher.ts
```diff
@@ -322,95 +328,52 @@ export class NTTWatcher extends Watcher {
...
+              );
+              const calculatedDigest = getNttManagerMessageDigest(
+                chainToChainId(this.chain),
+                transceiverMessage.ntt_managerPayload
+              );
+              const sourceToken: string =
+                transceiverMessage.ntt_managerPayload.payload.sourceToken.toString('hex');
+              const decimals = transceiverMessage.ntt_managerPayload.payload.trimmedAmount.decimals;
+              const amount = transceiverMessage.ntt_managerPayload.payload.trimmedAmount.amount;
+              const trimmedAmount: TrimmedAmount = new TrimmedAmount(amount, decimals);
```
bingyuyap, line 351: the `trimmedAmount` on the payload is already of `TrimmedAmount` class. So we can just do 

```ts
const normalizedAmount = transceiverMessage.ntt_managerPayload.payload.trimmedAmount.normalize(NTT_DECIMALS);
```

panoel: fixed.

bingyuyap (approved):
Thank you for making the changes!

PR #247, opened 2024-04-03
dashboard: cleanup deps

evan-gray:
Since `@certusone/wormhole-sdk` is no longer used, some dependencies were able to be cleaned up along with some unused ones.

This left `@cosmjs/cosmwasm-stargate` as the only dependency left requiring a config override with `react-app-rewired`. It was only used for `queryContractSmart`, so I simply implemented that with axios instead.

Overall, this about halved the bundle size down to `533.92 kB`

Also added `npm run analyze -w dashboard` for analyzing the webpack bundle. Seems like additional gains could be had by cleaning up the `@solana` dependencies from `common` or perhaps helping it tree-shake.


dashboard/src/utils/queryContractSmart.ts
```diff
@@ -0,0 +1,70 @@
...
+      prove: false,
+    },
+  });
+  if (!response.data.result.response.value) {
+    if (response.data.result.response.code && response.data.result.response.log) {
+      throw new Error(
+        `Query failed with (${response.data.result.response.code}) ${response.data.result.response.log}`
+      );
+    } else {
+      throw new Error(`Query failed with unknown error`);
```
bingyuyap, line 61: Do we also want to log something like `cosmwasm query` for easier debug?

bingyuyap (approved):
LGTM!

dashboard/src/utils/queryContractSmart.ts
evan-gray: I think it's obvious from the traceback. I was just trying to keep the error message similar to what the library was throwing.

PR #251, opened 2024-04-04
common: Magic number is 40?

panoel:


bingyuyap (approved):
You're a wizard, Paul.

PR #254, opened 2024-04-05
watcher: fix near

panoel:


watcher/src/watchers/NearArchiveWatcher.ts
```diff
@@ -151,4 +150,76 @@ export class NearArchiveWatcher extends Watcher {
...
+    let page = 1;
+    while (!done) {
+      console.log(`Fetching transactions for page ${page}...`);
+      // https://api3.nearblocks.io/v1/account/contract.wormhole_crypto.near/txns?method=publish_message&order=desc&page=1&per_page=25
+      const res = (
+        await axios.get(
+          `https://api3.nearblocks.io/v1/account/${accountId}/txns?method=publish_message&order=desc&page=${page}&per_page=25`,
+          AXIOS_CONFIG_JSON
+        )
+      ).data as GetTransactionsByAccountIdResponse;
```
evan-gray, line 172: you should use `axios.get<GetTransactionsByAccountIdResponse>` instead of `as`. IIRC the former enforces the type while the latter overrides it

watcher/src/watchers/NearArchiveWatcher.ts
```diff
@@ -151,4 +150,76 @@ export class NearArchiveWatcher extends Watcher {
...
+    console.log(`Fetching transactions in range [${beginningTimestamp}, ${endingTimestamp}]...`);
+    let txs: NearTxn[] = [];
+    let done: boolean = false;
+
+    let page = 1;
+    while (!done) {
+      console.log(`Fetching transactions for page ${page}...`);
+      // https://api3.nearblocks.io/v1/account/contract.wormhole_crypto.near/txns?method=publish_message&order=desc&page=1&per_page=25
+      const res = (
+        await axios.get(
```
bingyuyap, line 168: Do we also want to try catch this? On phone now so unsure if there is a higher level try catch block, sorry.

watcher/src/watchers/NearArchiveWatcher.ts
panoel: Fixed

bingyuyap (approved):
Thanks for making the changes!

PR #255, opened 2024-04-06
Bing/dashboard ntt capacity

bingyuyap:
This PR adds the inbound and outbound capacity for the supported chains onto the dashboard. 

1 issue:
- Outbound Capacity column shifts to the right very slightly, didn't fix because I don't to touch too much of the frontend code 

image


common/src/consts.ts
evan-gray, line 1: If these are only relevant to the dashboard, let’s keep them in the dashboard folder for now.

dashboard/package.json
```diff
@@ -17,7 +17,11 @@
     "@types/node": "^18.6.4",
     "@types/react": "^18.0.15",
     "@types/react-dom": "^18.0.6",
+    "@wormhole-foundation/sdk-definitions-ntt": "^0.0.1-beta.1",
+    "@wormhole-foundation/sdk-evm-ntt": "^0.0.1-beta.1",
+    "@wormhole-foundation/sdk-solana-ntt": "^0.0.1-beta.1",
     "axios": "^0.27.2",
+    "ethers": "^6.11.1",
```
evan-gray, line 24: Can we avoid the ethers dependency? We’ve done well avoiding it this far.

dashboard/src/components/LookerRateLimits.tsx
```diff
@@ -0,0 +1,101 @@
...
+      <Box textAlign="right">
+        {info.row.original.destChain
+          ? `$${info.row.original.amount.toLocaleString()}`
+          : `$${info.row.original.totalInboundCapacity?.toLocaleString()}`}
+      </Box>
+    ),
+  }),
+];
+
+export function LookerRateLimits() {
```
evan-gray, line 72: Why is this called “Looker” rate limits? It doesn’t have anything to do with Looker. Maybe NTTRateLimits would be more apt.

dashboard/src/hooks/useRateLimits.ts
```diff
@@ -0,0 +1,19 @@
...
+import { RateLimit, getRateLimits } from '../utils/nttHelpers';
+import { Network } from '../contexts/NetworkContext';
+
+export function useRateLimits(network: Network) {
+  const [rateLimits, setRateLimits] = useState<RateLimit[]>([]);
+
+  useEffect(() => {
+    const fetchRateLimits = async () => {
+      const limits = await getRateLimits(network);
+      setRateLimits(limits);
```
evan-gray, line 12: This effect does not handle when it’s cancelled which can lead to memory leaks (and a warning from react). See the other hooks for examples. Also probably worth auto refreshing every minute or so like the others.

evan-gray (changes requested):
Good start. In addition to the comments, one organizational thing seems incorrect - this will be used to monitor multiple networks of NTT tokens so the rate limits should probably be organized and displayed that way.

dashboard/package.json
bingyuyap: I'm adding the `tokenDecimal` call to the SDK, finding out whether Solana has the same call before doing that

dashboard/src/hooks/useRateLimits.ts
bingyuyap: I hope I did it right 🤣

dashboard/package.json
bingyuyap: removed, everything is axios now

dashboard/src/hooks/useRateLimits.ts
```diff
@@ -0,0 +1,36 @@
...
+  const [rateLimits, setRateLimits] = useState<RateLimit[]>([]);
+
+  useEffect(() => {
+    let cancelled = false;
+
+    const fetchRateLimits = async () => {
+      if (cancelled) return;
+      const limits = await getRateLimits(network);
+      limits.sort((a, b) => a.tokenName.localeCompare(b.tokenName));
+      setRateLimits(limits);
```
evan-gray, line 18: you should also check `cancelled` right before calling set since you're awaiting an async call above, it could have been cancelled between your initial check and here.

dashboard/src/components/NTTMetrics.tsx
```diff
@@ -21,6 +22,7 @@ function NTTMetrics() {
         title="Testnet NTT Transfers Report"
         src="https://lookerstudio.google.com/embed/reporting/a47057a8-15a0-4cc7-8086-eb00f5d09d2a/page/SPpuD"
       />
+      <NTTRateLimits />
```
panoel, line 25: Will you be adding NTTRateLimits to the mainnet metrics?

dashboard/src/utils/consts.ts
```diff
@@ -175,3 +176,61 @@ export const GUARDIAN_SET_3 = [
...
+      Sepolia: '0x1d30E78B7C7fbbcef87ae6e97B5389b2e470CA4a',
+      ArbitrumSepolia: '0x84A1Cb660B19eB0063EE5FD377eC14AAe3364d74',
+      BaseSepolia: '0x7f430D4e7939D994C0955A01FC75D9DE33F12D11',
+      OptimismSepolia: '0x0e15979a7a1eFAEf20312CA45A59eb141bF7E340',
+    },
+  },
+  Devnet: {},
+};
+
+export const NTT_MANAGER_CONTRACT: {
```
panoel, line 206: Aren't these in the NTTconsts.ts file?

dashboard/src/utils/consts.ts
```diff
@@ -175,3 +176,61 @@ export const GUARDIAN_SET_3 = [
...
+  Mainnet: [],
+  Devnet: [],
+};
+
+export type ContractAddresses = {
+  [key in Network]: { [key in Chain]?: string };
+};
+
+export const NTT_TOKENS: {
+  [key in Network]: { [tokenName: string]: { [key in Chain]?: string } };
```
panoel, line 191: Does it make sense to do a [tokenAddr, tokenName] here?  Will the tokenName be used outside of the embeded looker graphic?

dashboard/src/utils/consts.ts
panoel: Should we pull some of this into a common/NttConsts.ts file and pull the files out of watcher/ too?

dashboard/src/utils/evmNttHelpers.ts
```diff
@@ -0,0 +1,118 @@
...
+    // decimals should be uint8 so this should be safe to cast
+    return Number(result);
+  } catch (e) {
+    throw new Error(`Failed to get token decimals: ${e} for contract: ${contractAddress}`);
+  }
+}
+
+function getMethodId(methodSignature: string): string {
+  const hashArray = keccak256(methodSignature); // This is a Uint8Array
+
```
panoel, line 110:
```suggestion
  const hashArray: Uint8Array = keccak256(methodSignature);
```

dashboard/src/utils/evmNttHelpers.ts
```diff
@@ -0,0 +1,118 @@
...
+// function getCurrentOutboundCapacity() external view returns (uint256);
+export async function getCurrentOutboundCapacity(
+  rpc: string,
+  contractAddress: string
+): Promise<bigint> {
+  const data = getMethodId('getCurrentOutboundCapacity()');
+
+  // JSON-RPC request payload
+  const payload = {
```
panoel, line 12: Since this is common code for the next 3 functions, too bad you can't have a function that takes (rpc, contractAddr, data) and returns a response.

dashboard/src/components/NTTMetrics.tsx
bingyuyap: I should!

dashboard/src/hooks/useRateLimits.ts
bingyuyap: Will add

dashboard/src/utils/consts.ts
bingyuyap: Do you mean like this: 

```ts
export const NTT_TOKENS: TokenAddresses = {
  Mainnet: {},
  Testnet: {
    Solana: {
      '87r5ZS91Q2pQbFTvvneqs7y7mbtegtqMt4LDAS4g23Ax': 'TestToken',
    },
    Sepolia: {
      '0x1d30E78B7C7fbbcef87ae6e97B5389b2e470CA4a': 'TestToken',
    },
    ArbitrumSepolia: {
      '0x84A1Cb660B19eB0063EE5FD377eC14AAe3364d74': 'TestToken',
    },
    BaseSepolia: {
      '0x7f430D4e7939D994C0955A01FC75D9DE33F12D11': 'TestToken',
    },
    OptimismSepolia: {
      '0x0e15979a7a1eFAEf20312CA45A59eb141bF7E340': 'TestToken',
    },
  },
  Devnet: {},
};
```

The token name will be displayed in the table below the looker, if that's what you are asking about

dashboard/src/utils/consts.ts
bingyuyap: The `NTTConsts.ts` file supports a different structure, but I can update the logic to conform with that

dashboard/src/utils/evmNttHelpers.ts
bingyuyap: I'll try to modularize that logic

dashboard/src/utils/consts.ts
bingyuyap: Added the const file `common/src/nttConsts.ts` with a function to convert to the data structure that is usually used in the watchers

common/src/nttConsts.ts
```diff
@@ -0,0 +1,100 @@
...
+      OptimismSepolia: '0x0e15979a7a1eFAEf20312CA45A59eb141bF7E340',
+    },
+  },
+  Devnet: {},
+};
+
+export const NTT_MANAGER_CONTRACT_ARRAY =
+  convertNTTManagerContractToNTTContractArray(NTT_MANAGER_CONTRACT);
+
+export function SUPPORTED_CHAINS(network: Network, token: string): Chain[] {
```
evan-gray, line 98: this not being prefixed with `NTT_` makes me a little nervous given that this file is re-exported

common/src/nttConsts.ts
```diff
@@ -0,0 +1,100 @@
+import { Chain, Network } from '@wormhole-foundation/sdk-base';
+
+// This data structure is used in dashboard
+type NTTContract = {
```
evan-gray, line 4: should these types be exported?

common/src/nttConsts.ts
```diff
@@ -0,0 +1,100 @@
...
+    },
+  },
+  Devnet: {},
+};
+
+export const NTT_MANAGER_CONTRACT_ARRAY =
+  convertNTTManagerContractToNTTContractArray(NTT_MANAGER_CONTRACT);
+
+export function SUPPORTED_CHAINS(network: Network, token: string): Chain[] {
+  return Object.keys(NTT_MANAGER_CONTRACT[network][token]) as Chain[];
```
evan-gray, line 99: would be nice if there were a cleaner way to do this without the `as`

dashboard/src/components/NTTRateLimits.tsx
```diff
@@ -0,0 +1,134 @@
...
+              onClick: info.row.getToggleExpandedHandler(),
+            }}
+          >
+            {info.row.getIsExpanded() ? (
+              <KeyboardArrowDown fontSize="inherit" />
+            ) : (
+              <KeyboardArrowRight fontSize="inherit" />
+            )}
+          </IconButton>
+        ) : null}{' '}
```
evan-gray, line 69: feel like this could use an indent here for when there isn't a dropdown arrow, else the indentation is a bit odd

dashboard/src/utils/nttHelpers.ts
```diff
@@ -0,0 +1,163 @@
...
+  return rateLimitsByToken;
+}
+
+async function getEvmRateLimits(
+  network: Network,
+  chain: Chain,
+  tokenName: string,
+  contractAddress: string
+): Promise<RateLimit> {
+  const rpcAddress = rpc.rpcAddress(network.name as SdkNetwork, chain);
```
evan-gray, line 81: all of these rpc calls could have been batched into a single request. this might be fine for now but should be reconsidered as the list of tokens grows

watcher/src/watchers/NTTSolanaWatcher.ts
```diff
@@ -31,6 +30,7 @@ import {
   TransferLockIx,
   getNttManagerMessageDigest,
 } from '../NTTConsts';
+import { NTT_MANAGER_CONTRACT_ARRAY } from '../../../common/src/nttConsts';
```
evan-gray, line 33: this import should be from the package, not the file path

watcher/src/watchers/NTTWatcher.ts
```diff
@@ -25,6 +24,7 @@ import {
   createNewLifeCycle,
   getNttManagerMessageDigest,
 } from '../NTTConsts';
+import { NTT_MANAGER_CONTRACT_ARRAY } from '../../../common/src/nttConsts';
```
evan-gray, line 27: this import should be from the package, not the file path

dashboard/src/utils/nttHelpers.ts
```diff
@@ -0,0 +1,163 @@
...
+  Network as SdkNetwork,
+  chainToChainId,
+  rpc,
+} from '@wormhole-foundation/sdk-base';
+import { Network } from '../contexts/NetworkContext';
+import {
+  NTT_MANAGER_CONTRACT,
+  NTT_TOKENS,
+  SUPPORTED_CHAINS,
+} from '@wormhole-foundation/wormhole-monitor-common/dist/nttConsts';
```
evan-gray, line 13: should this use the top level export instead?

evan-gray (changes requested):

common/src/nttConsts.ts
bingyuyap: I think `Object.keys` inherently return `string[]`. Another way would be something like

```ts
export function NTT_SUPPORTED_CHAINS(network: Network, token: string): Chain[] {
  const contractDetails = NTT_MANAGER_CONTRACT[network][token];
  if (!contractDetails) {
    return [];
  }

// chains imported from sdk-base
  return chains.filter((chain) => chain in contractDetails);
}
```

dashboard/src/utils/nttHelpers.ts
bingyuyap: let's do this in the coming pull request

dashboard/src/components/NTTRateLimits.tsx
bingyuyap: added!

watcher/src/watchers/NTTSolanaWatcher.ts
bingyuyap: changed

watcher/src/watchers/NTTWatcher.ts
bingyuyap: changed

dashboard/src/utils/nttHelpers.ts
bingyuyap: changed

evan-gray (approved):

PR #256, opened 2024-04-08
watcher: add NTT contracts to mainnet

panoel:


watcher/src/NTTConsts.ts
```diff
@@ -154,7 +154,10 @@ export const NTT_SOLANA_IXS = [
 ];
 
 export const NTT_CONTRACT: { [key in Network]: { [key in Chain]?: string[] } } = {
-  ['Mainnet']: {},
+  ['Mainnet']: {
+    Ethereum: ['0x4d573bc8cE236BE2609333206776c5B6FB8f4a10'],
```
bingyuyap, line 158: These are Implementation Contracts. Correct me if I'm wrong, but  I'm wondering if we should use the Proxy Contracts instead. I think when the contracts are upgraded the Proxy Contracts will point to a new Implementation.

panoel: I have no idea.  Who knows this answer?

panoel: Fixing

panoel: So, the events are emitted by the proxy and not the actual manager contract?

bingyuyap: For wormhole at least events are emitter from proxy, I just confirmed with Karl. I think it might be the way `delegateCall` works, it's essentially taking the impl contract code and executing that in the proxy contract itself. This way the storage and context are still the proxy's.

image
found this image from this [page](https://medium.com/@ajaotosinserah/mastering-delegatecall-in-solidity-a-comprehensive-guide-with-evm-walkthrough-6ddf027175c7)

Etherscan also shows the proxy emitting the events as well: https://etherscan.io/address/0xebdce9a913d9400ee75ef31ce8bd34462d01a1c1#events

bingyuyap (approved):

PR #262, opened 2024-04-12
cloud_functions: migratiing to sdk V2

panoel:


cloud_functions/src/computeMessageCounts.ts
```diff
@@ -53,7 +53,7 @@ async function getMessageCounts_(prevMessageCounts: CountsByChain): Promise<Coun
   const [observedMessages] = await table.getRows(ranges);
   const [missingVaaObservedMessages] = await table.getRows(missingVaaRanges);
 
-  for (const [chainName, chainId] of Object.entries(CHAINS)) {
+  for (const chainId of chainIds) {
```
bingyuyap, line 56: does this matter if it loops through testnet chain ids as well?

cloud_functions/src/computeMessageCountHistory.ts
```diff
@@ -39,16 +40,16 @@ export async function computeMessageCountHistory(req: any, res: any) {
...
         if (parsed.timestamp === 0) {
           // e.g. governance VAAs may have timestamp set to 0
           continue;
         }
         const date = new Date(parsed.timestamp * 1000).toISOString().slice(0, 10);
         messageCounts.DailyTotals[date] = {
           ...messageCounts.DailyTotals[date],
-          [parsed.emitterChain]:
-            (messageCounts.DailyTotals[date]?.[parsed.emitterChain.toString()] || 0) + 1,
+          [toChainId(parsed.emitterChain)]:
```
bingyuyap, line 51: Is the practice to use `chainToChainId` or `toChainId`? Not sure if that matters but I think `chainToChainId` is more type safe as it ensures that only valid chain names can be passed in. 

deferring this to @barnjamin

cloud_functions/src/getMessages.ts
```diff
@@ -235,8 +223,8 @@ export async function getMessages(req: any, res: any) {
     chain = pathVars[1];
     console.log(chain);
     let chainId: ChainId;
-    if (validChain(chain)) {
-      chainId = Number(chain) as ChainId;
+    if (isChainId(Number(chain))) {
```
bingyuyap, line 226: I am unsure if `Number(chain)` works if chain is a string, instead we can do `isChain(chain)` instead. It is provided in `sdk-base`

cloud_functions/src/types.ts
```diff
@@ -45,11 +45,6 @@ export type MessagesByChain = {
 
 export function makeCache() {
```
bingyuyap, line 46: not really related to this PR but maybe we can use this to prevent hardcoding

```ts
const cache: MessagesByChain = {};

for (const chainId of ChainIds) { // from sdk
  cache[chainId] = {
    messages: [],
    lastUpdated: 0,
    lastRowKey: '',
  };
}

return cache; 
```

cloud_functions/src/updateTokenMetadata.ts
```diff
@@ -77,19 +77,20 @@ export async function updateTokenMetadata(req: any, res: any) {
       symbol,
       decimals,
     } of result) {
-      assertChain(token_chain);
+      assertChainId(token_chain);
+      const tokenChain: ChainId = toChainId(token_chain);
```
bingyuyap, line 81: nit: `tokenChain` -> `tokenChainId`?

bingyuyap (dismissed):
Appreciate the work! 

Unsure about the practice about `toChainId` vs `chainToChainId`, not sure if it even matters. cc @barnjamin

cloud_functions/src/computeMessageCounts.ts
panoel: That is what it did before.  So, shouldn't be a concern.

cloud_functions/src/computeMessageCountHistory.ts
panoel: Changed.

cloud_functions/src/getMessages.ts
panoel: You are correct.  Fixed.

cloud_functions/src/types.ts
panoel: I'm struggling to get this code to work in a reasonable way.  IF you can, commit the change.

cloud_functions/src/updateTokenMetadata.ts
panoel: Changed.

common/src/utils.ts
```diff
@@ -38,3 +39,21 @@ export function getMode(): Mode {
...
 }
+
+// This function basically strips off the `0x` prefix from the hex string.
+export function universalAddress_stripped(u: UniversalAddress): string {
+  return encoding.hex.encode(u.toUint8Array());
+}
+
+// This function takes a Chain or a ChainId as a string and returns
+// the corresponding ChainId (or undefined if the chain is not recognized).
+export function stringToChainId(input: string): ChainId | undefined {
```
bingyuyap, line 50: I think you might be replicating what the sdk is doing under the hood on `toChainId()`. Check out the implementation [here](https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/core/base/src/constants/chains.ts#L92)

```ts
export const toChainId = (chain: number | string): ChainId => {
  switch (typeof chain) {
    case "string":
      if (isChain(chain)) return chainToChainId(chain);
      break;
    case "number":
      if (isChainId(chain)) return chain;
      break;
  }
  throw Error(`Cannot convert to ChainId: ${chain}`);
};
```

cloud_functions/src/consts.ts
```diff
@@ -1,8 +1,8 @@
-import { ChainId, CHAIN_ID_FANTOM } from '@certusone/wormhole-sdk';
+import { ChainId, toChainId } from '@wormhole-foundation/sdk-base';
 
 // TODO: this should probably be a table in the database
 export const TVL_TOKEN_DENYLIST: { [key in ChainId]?: string[] } = {
-  [CHAIN_ID_FANTOM]: ['0x5b2af7fd27e2ea14945c82dd254c79d3ed34685e'], // coingecko reporting bad prices
+  [toChainId('Fantom')]: ['0x5b2af7fd27e2ea14945c82dd254c79d3ed34685e'], // coingecko reporting bad prices
```
bingyuyap, line 5: In this case `Fantom` is in the Chain list so it would make sense to use `chainToChainId` for type safety, no harm doing `toChainId` though

cloud_functions/src/computeMessageCountHistory.ts
```diff
@@ -39,16 +40,20 @@ export async function computeMessageCountHistory(req: any, res: any) {
...
           continue;
         }
-        const parsed = parseVaa(signedVAA.data.info.bytes[0].value);
+        const parsed = deserialize('Uint8Array', signedVAA.data.info.bytes[0].value);
         if (parsed.timestamp === 0) {
           // e.g. governance VAAs may have timestamp set to 0
           continue;
         }
         const date = new Date(parsed.timestamp * 1000).toISOString().slice(0, 10);
+        const chainId = stringToChainId(parsed.emitterChain);
```
bingyuyap, line 49: `parsed.emitterChain` is of the Chain list type as well, so `chainToChainId` works here and we know that it shouldn't throw error

cloud_functions/src/getMessages.ts
```diff
@@ -234,10 +223,8 @@ export async function getMessages(req: any, res: any) {
   if (pathVars) {
     chain = pathVars[1];
     console.log(chain);
-    let chainId: ChainId;
-    if (validChain(chain)) {
-      chainId = Number(chain) as ChainId;
-
+    const chainId: ChainId | undefined = stringToChainId(chain);
```
bingyuyap, line 226: this makes sense to use `toChainId` since chain is a string type and we do not know if it lies within the Chain params. We want it to throw when its not a valid chain

cloud_functions/src/getReobserveVaas.ts
```diff
@@ -115,7 +114,7 @@ async function getAndProcessReobsVAAs(): Promise<Map<string, ReobserveInfo>> {
         // So we put the rest back in the collection
         let solanaCount = 0;
         for (const vaa of vaas) {
-          if (vaa.chain === CHAIN_ID_SOLANA) {
+          if (vaa.chain === toChainId('Solana')) {
```
bingyuyap, line 117: Similar to `Fantom` case, `chainToChainId` makes more sense to me here since `Solana` is definitely a valid Chain

cloud_functions/src/getReobserveVaas.ts
```diff
@@ -144,7 +143,7 @@ async function getAndProcessReobsVAAs(): Promise<Map<string, ReobserveInfo>> {
 async function processVaa(vaa: ReobserveInfo): Promise<ReobserveInfo[] | null> {
   let vaas: ReobserveInfo[] = [];
 
-  if (vaa.chain === CHAIN_ID_SOLANA) {
+  if (vaa.chain === toChainId('Solana')) {
```
bingyuyap, line 146: Here as well

common/src/utils.ts
panoel: That is not correct.  If you pass a number in quotes (eg '1') to the sdk, it will throw.  It only supports an actual number (Number type) or the quoted chain name (string type).

bingyuyap: You are right, but I wonder if we every have the case where number in quotes are introduced in the codebase, maybe when reading from json?

cloud_functions/src/computeMessageCountHistory.ts
panoel: Changed

cloud_functions/src/getMessages.ts
panoel: `chain` here is a quoted number.  `toChainId` will always throw in this case.  Also, I believe the intent of the original code was to skip any anomalous chainId and return the good data.  Throwing here likely does not adhere to the original intent.

common/src/utils.ts
panoel: See getMessages.ts

cloud_functions/src/consts.ts
panoel: Changed.

cloud_functions/src/getReobserveVaas.ts
panoel: Changed.

cloud_functions/src/getMessages.ts
bingyuyap: That makes sense!

bingyuyap (approved):
LGTM!

PR #263, opened 2024-04-15
cloud_function: get ntt rate limits

bingyuyap:
this PR depends on https://github.com/wormhole-foundation/wormhole-dashboard/pull/255


cloud_functions/src/getNTTRateLimits.ts
```diff
@@ -0,0 +1,130 @@
...
+  let ntt: EvmNtt<Network, EvmChains> | SolanaNtt<Network, 'Solana'>;
+  let tokenDecimals: number;
+  const rpcEndpoint = rpc.rpcAddress(network, chain as Chain);
+  const tokenAddress = NTT_TOKENS[network][token][chain]!;
+  const managerContract = NTT_MANAGER_CONTRACT[network][token][chain]!;
+  const transceiverContract = NTT_TRANSCEIVER_CONTRACT[network][token][chain]!;
+
+  if (isSolana) {
+    const platform = new SolanaPlatform(network);
+    ntt = new SolanaNtt(network, 'Solana', platform.getRpc('Solana'), {
```
panoel, line 49: I think it is silly that if you are creating something called 'SolanaNtt' that you have to pass in 'Solana'.  But this is out of your control, right?

bingyuyap: SolanaNtt takes in either Solana or Pythnet as a param so in this case it cant be helped

cloud_functions/src/getNTTRateLimits.ts
```diff
@@ -0,0 +1,130 @@
...
+
+async function getRateLimits(network: Network, token: string, chain: Chain, isSolana: boolean) {
+  let ntt: EvmNtt<Network, EvmChains> | SolanaNtt<Network, 'Solana'>;
+  let tokenDecimals: number;
+  const rpcEndpoint = rpc.rpcAddress(network, chain as Chain);
+  const tokenAddress = NTT_TOKENS[network][token][chain]!;
+  const managerContract = NTT_MANAGER_CONTRACT[network][token][chain]!;
+  const transceiverContract = NTT_TRANSCEIVER_CONTRACT[network][token][chain]!;
+
+  if (isSolana) {
```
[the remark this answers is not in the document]
bingyuyap: that's true

cloud_functions/src/getNTTRateLimits.ts
```diff
@@ -0,0 +1,130 @@
...
+    { token, chain, network, product: PRODUCT },
+    normalizedOutboundCapacity
+  );
+
+  const inboundChains = NTT_SUPPORTED_CHAINS(network, token).filter(
+    (inboundChain) => inboundChain !== chain
+  );
+  await Promise.all(
+    inboundChains.map(async (inboundChain) => {
+      const inboundCapacity = await ntt.getCurrentInboundCapacity(inboundChain);
```
panoel, line 89: Should we be summing these all up?  Am I missing that?  Looks like gauge will quickly run through all the specific inbound capacities and then stay on the last value until the next cycle?

panoel: Is there one gauge per inbound chain?

bingyuyap: I don't think there is a need to sum these up, as the `inbound_chain` detail of `inbound_capacity` is still important. If we want to we can quickly sum them up by `chain` on Grafana. 

What we are doing is to use the prometheus to scrape this cloud_function every interval. What we want to know is the current inbound/outbound capacity for each chain. So this is just putting the current capacity into the gauge and returning the data to prometheus. 

There is one gauge metric per `inbound_chain`. EG:
```
ntt_inbound_capacity{token="TEST_NTT",chain="ArbitrumSepolia",inbound_chain="Solana",network="Testnet",product="cloud_functions_ntt"} 25
ntt_inbound_capacity{token="TEST_NTT",chain="ArbitrumSepolia",inbound_chain="Sepolia",network="Testnet",product="cloud_functions_ntt"} 999999
ntt_inbound_capacity{token="TEST_NTT",chain="ArbitrumSepolia",inbound_chain="BaseSepolia",network="Testnet",product="cloud_functions_ntt"} 999999
ntt_inbound_capacity{token="TEST_NTT",chain="ArbitrumSepolia",inbound_chain="OptimismSepolia",network="Testnet",product="cloud_functions_ntt"} 999999
```

panoel (commented):
Please have @evan look at the dashboard components.  I'm not fluent in that part.

panoel (dismissed):

common/src/evm.ts
```diff
@@ -33,36 +33,22 @@ async function callContractMethod(
...
-  return BigInt(result);
+  return Number(result);
 }
 
-export async function getTokenDecimals(rpc: string, contractAddress: string): Promise<number> {
-  const methodId = getMethodId('tokenDecimals()');
+export async function getEvmTotalSupply(rpc: string, contractAddress: string): Promise<number> {
+  const methodId = getMethodId('totalSupply()');
   const result = await callContractMethod(rpc, contractAddress, methodId);
   return Number(result);
```
evan-gray, line 53: This doesn't seem safe to be a `Number`, `BigInt` is likely better suited.

common/src/solana.ts
```diff
@@ -3,8 +3,12 @@ import {
   Message,
   MessageCompiledInstruction,
   MessageV0,
+  PublicKeyInitData,
+  PublicKey,
 } from '@solana/web3.js'; // NOTE: types only for bundling size
```
evan-gray, line 8: See this comment. Prior to this PR, only the types from `@solana/web3.js` were used. Now this is no longer the case. Overall the changes appear to have an ~11% increase on the bundle size (421.12 kB (+39.53 kB)). I am open to keeping this change as is for now, but the comment should be removed since it's no longer true.

dashboard/src/components/NTTMetrics.tsx
```diff
@@ -12,7 +15,40 @@ function NTTMetrics() {
           src="https://lookerstudio.google.com/embed/reporting/0f20bce5-d442-4f39-8cc4-ced8bb73042a/page/kSKuD"
           hasTabs
         />
-        <NTTRateLimits />
+        <Divider />
+        <CollapsibleSection
```
evan-gray, line 19: Looks like the other sections typically put the `CollapsibleSection` inside the component so the higher level component is cleaner.

evan-gray (approved):
Reviewed `common` and `dashboard`. Might be worth adding some loaders for the tables, at least the first time I hit them it was a little slow. Can be done in a follow up though.

PR #264, opened 2024-04-16
dashboard: ntt total supply on evm vs total locked on sol

bingyuyap:
This PR does a few things:
- moves `ntt_rate_limit` and `ntt_total_supply_locked` rpc invocation from frontend to cloud function due to Solana Mainnet RPC CORS gating
  - `computeNTT<something>`: invokes RPC and saves data into a bucket cache
  - `getNTT<something>`: takes data from bucket cache and return to client
- getNTTRateLimits returns 2 types of format to minimize code logic duplication:
  - Prometheus text format by default
  - JSON - if request `Content-Type` is `application/json`
- Note that `getNTTRateLimits` is gated by a fake bearer token because Grafana Metrics Endpoint Integration mandates the URL to be gated
- adds a table to show number of tokens locked on Solana vs total number of tokens on EVM chains

image


common/src/solana.ts
```diff
@@ -50,3 +53,64 @@ export const getSolanaTokenDecimals = async (rpc: string, mintAddress: string):
...
+  if (response.data.error) {
+    throw new Error(`Error fetching ${method} account: ${response.data.error.message}`);
+  }
+  return response.data.result;
+}
+
+// Helper function to decode base64 data
+export function decodeBase64Data(encodedData: string) {
+  return new Uint8Array(
+    atob(encodedData)
```
panoel, line 112: This function is deprecated.  Can we not use Buffer.from(encodedData, "base64"), or something like that, instead?

cloud_functions/src/computeNTTRateLimits.ts
```diff
@@ -0,0 +1,133 @@
...
+
+        const inboundCapacity = await Promise.all(inboundCapacityPromises);
+
+        return {
+          tokenName: token,
+          inboundCapacity: inboundCapacity,
+        };
+      })
+    );
+    await Promise.all(rateLimitPromises);
```
panoel, line 126: Do you still have to do this await even though there is an await on line 110?

cloud_functions/src/computeNTTRateLimits.ts
```diff
@@ -0,0 +1,133 @@
...
+let bucketName: string = 'wormhole-ntt-cache';
+if (assertEnvironmentVariable('NETWORK') === 'Testnet') {
+  bucketName = 'wormhole-ntt-cache-testnet';
+}
+
+const cacheBucket = storage.bucket(bucketName);
+const cacheFileName = 'ntt-rate-limits-cache.json';
+const cloudStorageCache = cacheBucket.file(cacheFileName);
+
+async function computeNTTRateLimits_(network: Network, token: string, chain: Chain) {
```
panoel, line 28: Typescript question: does this function need to return Promise<void> ?

panoel (commented):
I did not review the components files as I'm not sure how they work.

common/src/solana.ts
bingyuyap: Updated

cloud_functions/src/computeNTTRateLimits.ts
bingyuyap: You are right! Changed this

cloud_functions/src/computeNTTRateLimits.ts
bingyuyap: by default async functions return Promises, but I will add a Promise<NTTRateLimit> for enforcing type checking

panoel (approved):

PR #269, opened 2024-04-18
watcher: update NTT quoter contracts

panoel:


evan-gray (approved):
These seem like quoters insofar as I can tell

PR #272, opened 2024-04-23
ft_watcher: initial bootstrap for fast transfer watcher

bingyuyap:
PR does few things:
1. Fast Transfer Matching Engine ix parsing - `placeInitialOfferCctp` and `improveOffer`
    - this allows us to at least track time taken for first offer to occur. 
2. Tests to verify ix parsing
3. add psql schema for tracking fast transfers and auctions
4. copy pasta of https://github.com/wormhole-foundation/example-liquidity-layer/tree/main/solana/ts/src in `watcher/src/fastTransfer` as a **temporary dependency until sdk is out** (review not needed here)
    - there are quite a few dependencies here so would take a bit of time to trim it down to what we actually need
    - due to the constant updates from `example-liquidity-layer`, it is easier to just copy pasta
    
 Note: PR is not feature ready but at a good checkpoint for review to prevent bulky PRs in the future
 Also the actual PR is 442 lines, everything else is from `example-liquidity-layer` and IDLs

Don't need to review: 
```
watcher/src/fastTransfer
watcher/src/idls
watcher/src/types
```


database/fast-transfer-schema.sql
```diff
@@ -0,0 +1,25 @@
...
+  auction_pubkey VARCHAR(255),
+  amount BIGINT NOT NULL,
+  initial_offer_time TIMESTAMP,
+  src_chain INTEGER NOT NULL,
+  dst_chain INTEGER NOT NULL,
+  sender VARCHAR(255) NOT NULL,
+  redeemer VARCHAR(255) NOT NULL,
+  tx_hash VARCHAR(255) NOT NULL,
+  timestamp TIMESTAMP NOT NULL
+);
```
panoel, line 13: Did you want to add a PRIMARY KEY to these tables?

watcher/src/watchers/FastTransferSolanaWatcher.ts
```diff
@@ -0,0 +1,321 @@
...
+import MATCHING_ENGINE_IDL from '../idls/matching_engine.json';
+import TOKEN_ROUTER_IDL from '../idls/token_router.json';
+import { RPCS_BY_CHAIN } from '../consts';
+import { VaaAccount } from '../fastTransfer/wormhole';
+import { LiquidityLayerMessage } from '../fastTransfer/common';
+import { AuctionOffer, FastTransfer, ParsedLogs } from '../fastTransfer/types';
+import knex, { Knex } from 'knex';
+import { assertEnvironmentVariable } from '@wormhole-foundation/wormhole-monitor-common';
+
+// TODO: this is devnet consts
```
panoel, line 23: Did you need to convert this to something like `MATCHING_ENGINE_PROGRAM_ID[network: Network]`?

panoel: Same for the 2 following consts

watcher/src/watchers/FastTransferSolanaWatcher.ts
```diff
@@ -0,0 +1,321 @@
...
+  readonly rpc: string;
+  readonly matchingEngineBorshCoder: BorshCoder;
+  readonly tokenRouterBorshCoder: BorshCoder;
+  readonly matchingEngineProgram: MatchingEngineProgram;
+  readonly pg: Knex | null = null;
+
+  constructor(network: Network, isTest: boolean = false) {
+    super(network, false);
+
+    this.rpc = isTest ? 'https://api.devnet.solana.com' : RPCS_BY_CHAIN[this.network].Solana!;
```
panoel, line 38: Can you derive `isTest` from the network?

watcher/src/watchers/FastTransferSolanaWatcher.ts
```diff
@@ -0,0 +1,321 @@
...
+          password: assertEnvironmentVariable('PG_NTT_PASSWORD'),
+          database: assertEnvironmentVariable('PG_NTT_DATABASE'),
+          host: assertEnvironmentVariable('PG_NTT_HOST'),
+          port: Number(assertEnvironmentVariable('PG_NTT_PORT')),
+        },
+      });
+    }
+  }
+
+  // TODO: Modify this so watcher can actually call this function (Add enum for mode)
```
panoel, line 64: I don't understand this todo

watcher/src/watchers/FastTransferSolanaWatcher.ts
```diff
@@ -0,0 +1,321 @@
...
+  async getMessagesByBlock(fromSlot: number, toSlot: number): Promise<string> {
+    this.logger.info(`fetching info for blocks ${fromSlot} to ${toSlot}`);
```
panoel, line 66: Did you want to check range here?
```
if (fromSlot > toSlot) throw new Error('solana: invalid block range');
```

watcher/src/watchers/FastTransferSolanaWatcher.ts
```diff
@@ -0,0 +1,321 @@
...
+    // We want to sort them by chronological order and process them from the oldest to the newest
+    const results = await this.getConnection().getTransactions(
+      signatures.map((s) => s.signature),
+      {
+        maxSupportedTransactionVersion: 0,
+      }
+    );
+
+    if (results.length !== signatures.length) {
+      throw new Error(`solana: failed to fetch tx for signatures`);
```
panoel, line 136: Did you want prepend your error messages with `fast transfer solana` or `FT solana` instead of `solana`?

watcher/src/watchers/FastTransferSolanaWatcher.ts
```diff
@@ -0,0 +1,321 @@
...
+      //   currSignature,
+      //   toSignature,
+      //   new PublicKey(TOKEN_ROUTER_PROGRAM_ID)
+      // );
+
+      numSignatures = signatures.length;
+      // We can safely use the last signature from either program as the loop continuation
+      // `fetchAndProcessMessages` will check every signature between `currSignature` and `toSignature` for both programs
+      // hence no signature will be missed
+      // TODO: We should check which one is the later signature and use that as the continuation to reduce duplicated processing
```
panoel, line 95: Did you want to create an issue for this, if you aren't solving this in this PR?

database/fast-transfer-schema.sql
bingyuyap: Yes, `fast_transfer_id` for `fast_transfers` table so we make sure we track only 1 row for 1 transfer. `tx_hash` for `auction_logs` since placing offers for an auction are in their own tx

watcher/src/watchers/FastTransferSolanaWatcher.ts
bingyuyap: Yep, will do this in the next PR

watcher/src/watchers/FastTransferSolanaWatcher.ts
bingyuyap: No, `isTest` and `network`  are independent. The test files do `Testnet` and `isTest = true` so it tests on testnet data and not require the psql database.

watcher/src/watchers/FastTransferSolanaWatcher.ts
bingyuyap: Current watcher has a `isNTT` boolean to check for:
- initial and latest block to use 
- which `getMessagesForBlocks`  to use 
- what data to store in firestore 

this current FT watcher wouldn't work in this logic so my thinking is to add an enum `normal`, `ntt`, `ft` to this logic to plug the ft watcher in

watcher/src/watchers/FastTransferSolanaWatcher.ts
```diff
@@ -0,0 +1,321 @@
...
+    } else if (programId.toBase58() === TOKEN_ROUTER_PROGRAM_ID) {
+      decodedData = this.tokenRouterBorshCoder.instruction.decode(
+        Buffer.from(instruction.data),
+        'base58'
+      );
+    }
+
+    const ixName = decodedData?.name;
+    if (!ixName) {
+      this.logger.debug('decodedData is null');
```
panoel, line 192: This is expected from time-to-time and is not an error condition?

watcher/src/watchers/FastTransferSolanaWatcher.ts
```diff
@@ -0,0 +1,321 @@
...
+
+    return { fast_transfer: null, auction_offer };
+  }
+
+  async saveFastTransfer(fastTransfer: FastTransfer): Promise<void> {
+    // this is to allow ci to run without a db
+    if (!this.pg) {
+      return;
+    }
+    this.logger.info('saving fast transfer');
```
panoel, line 306: Maybe add the fast transfer id to the message?  Otherwise, make it a debug.

watcher/src/watchers/FastTransferSolanaWatcher.ts
```diff
@@ -0,0 +1,321 @@
...
+    // Upsert the fast transfer
+    await this.pg('fast_transfers').insert(fastTransfer).onConflict('fast_transfer_id').merge();
+  }
+
+  async saveAuctionLogs(auctionLogs: AuctionOffer): Promise<void> {
+    // this is to allow ci to run without a db
+    if (!this.pg) {
+      return;
+    }
+    this.logger.info('saving auction logs');
```
panoel, line 317: Maybe add the fast VAA hash to the message? Otherwise, make it a debug.

watcher/src/watchers/FastTransferSolanaWatcher.ts
bingyuyap: This should happen when the instruction layout does not match the IDL. I will return null if program id is not what we want to look at

watcher/src/watchers/FastTransferSolanaWatcher.ts
bingyuyap: updated

watcher/src/watchers/FastTransferSolanaWatcher.ts
bingyuyap: updated

watcher/src/watchers/FastTransferSolanaWatcher.ts
bingyuyap: will add the hash but also make it a debug, there will be a lot of inserts so want to reduce logging

panoel (approved):

database/fast-transfer-schema.sql
panoel: Right.  So, do you need to add 
```
PRIMARY KEY (fast_transfer_id)
```
to the sql?

bingyuyap: it's already added here
```
fast_transfer_id VARCHAR(255) PRIMARY KEY,
```
which should work since it's only 1 primary key

panoel: Sorry, missed it!

PR #277, opened 2024-04-24
follow-up for #264 comments

bingyuyap:
dashboard: move CollapsibleSection into component
cloud_functions: use bignumber to handle bigint


cloud_functions/src/computeTotalSupplyAndLocked.ts
```diff
@@ -39,7 +40,10 @@ async function getEvmNormalizedTotalSupply(
     NTT_TOKENS[network][token][chain]!
   );
 
-  return tokenSupply / 10 ** tokenDecimals;
+  const base = new BigNumber(10).exponentiatedBy(tokenDecimals);
```
evan-gray, line 43: Why add a BigNumber dependency here when BigInt is built in and likely suitable?

bingyuyap: In this case normalizing the token supply by token decimals will lead to decimal points but bigint doesn't not support decimals, hence the decision of using BigNumber

evan-gray: these should be tracked separately - returning the amount as a `string` since it will be sent as JSON (using BigInt as the intermediate value) and the decimals separate as a number since it is small enough to be safe (or string). any amounts that are not safe as numbers (amounts, total supply, etc) should be `bigint`s and sent in JSON as strings (**not `Number`s**)

bingyuyap: updated!

evan-gray (dismissed):

dashboard/src/components/NTTTotalSupplyAndLocked.tsx
```diff
@@ -53,13 +56,23 @@ const rateLimitColumns = [
     header: () => <Box order="1">Locked</Box>,
     cell: (info) => (
       <Box textAlign="right">
-        {info.row.original.evmTotalSupply ? info.row.original.amountLocked?.toLocaleString() : null}
+        {info.row.original.evmTotalSupply
+          ? BigNumber(normalizeBigNumber(info.row.original.amountLocked, 2))
```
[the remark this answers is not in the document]
bingyuyap: You are right. I changed it. Sorry for the oversight

evan-gray (dismissed):

evan-gray (approved):

PR #279, opened 2024-04-26
cloud_function: makeCache to build from sdk chains

bingyuyap:
dynamically build cache using chains from sdkv2 instead of hardcoding

tested using 
```
{{cloud_function}}/messages/1
{{cloud_function}}/messages/1?reloadCache=true
{{cloud_function}}/messages/10003?reloadCache=true
```


cloud_functions/src/types.ts
```diff
@@ -43,250 +43,18 @@ export type MessagesByChain = {
...
       messages: [],
       lastUpdated: 0,
       lastRowKey: '',
-    },
-  };
-  return cache;
+    };
+  }
+
+  return cache as MessagesByChain;
```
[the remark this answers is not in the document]
evan-gray: upon further review, I think the type could be changed to `[chain in ChainId]?` akin to the other returns - this may even be safer for consumers with mismatched sdk versions

evan-gray (approved):

cloud_functions/src/getMessages.ts
```diff
@@ -234,26 +234,33 @@ export async function getMessages(req: any, res: any) {
...
-        if (
-          noVaaCache[chainId]['messages'].length === 0 ||
-          Date.now() - noVaaCache[chainId]['lastUpdated'] > REFRESH_TIME_INTERVAL
-        ) {
-          if (noVaaCache[chainId]['messages'].length === 0) {
-            console.log(`noVaaCache is empty, setting cache['messages] ${new Date()}`);
+        if (!noVaaCache[chainId]) {
+          noVaaCache[chainId] = {
+            messages: [],
+            lastUpdated: Date.now(),
```
evan-gray, line 240: wonder if this should be 0 or share code with `makeCache`, like a function for `makeCacheEntry` or something that returns this new object.

bingyuyap: good catch, added `makeCacheEntry` and used `0`

evan-gray (approved):

PR #280, opened 2024-04-26
common/cloud_functions/watcher: add scroll and blast

panoel:


evan-gray (changes requested):
I think this should also update `TOKEN_BRIDGE_EMITTERS` (and subsequently push an updated cloudfunction so that no transfers are missed)

evan-gray (approved):

PR #283, opened 2024-05-02
common: explorer links for new chains

evan-gray:


panoel (approved):
Sorry, I should have caught this one earlier.

PR #284, opened 2024-05-02
dashboard: override NTT token address to map to metadata

evan-gray:


bruce-riley (approved):

dashboard/src/components/Accountant.tsx
```diff
@@ -47,6 +47,15 @@ import CollapsibleSection from './CollapsibleSection';
 import { ExplorerTxHash } from './ExplorerTxHash';
 import Table from './Table';
 
+const NTT_ACCOUNTANT_TOKEN_ADDRESS_OVERRIDE: {
+  [chain: number]: { [tokenAddress: string]: string };
```
panoel, line 51: Should chain be a chainId instead of a number?

dashboard/src/components/Accountant.tsx
```diff
@@ -382,10 +391,12 @@ const MemoizedAccountantSearch = memo(AccountantSearch);
...
 function Accountant({
   governorInfo,
+  tokenData,
   accountantAddress,
   isNTT,
 }: {
   governorInfo?: CloudGovernorInfo;
+  tokenData: TokenDataByChainAddress | null;
   accountantAddress: string;
   isNTT?: boolean;
```
panoel, line 401: Does `isNTT` need to be both passed in and returned?  Seems unnecessary?

dashboard/src/components/Accountant.tsx
evan-gray: I just kept the type the same as where it's used - otherwise there'd be a typing issue. number should be fine for here.

dashboard/src/components/Accountant.tsx
evan-gray: returned? not sure I follow. this is the type definition of the parameters

panoel: Oh, oops.

PR #290, opened 2024-05-08
dashboard: switch to vite

evan-gray:
This PR switches out `react-scripts` for `vite` along with what I understand to be the required changes. The bundle size did appear to grow a bit (~50k) though that may be due to different handling of the wormhole sdk deps. Still under 500k gzip, but barely.


panoel:
Do we care about this:
```
vite v5.2.11 building for production...
../node_modules/@scure/base/lib/esm/index.js (350:33): A comment

"/* @__PURE__ */"

in "../node_modules/@scure/base/lib/esm/index.js" contains an annotation that Rollup cannot interpret due to the position of the comment. The comment will be removed to avoid issues.
../node_modules/@protobufjs/inquire/index.js (12:18): Use of eval in "../node_modules/@protobufjs/inquire/index.js" is strongly discouraged as it poses security risks and may cause issues with minification.
✓ 12388 modules transformed.
```

panoel (dismissed):

panoel (approved):

PR #292, opened 2024-05-09
Config page

evan-gray:
Adds a new page to display core contract info starting with the current guardian set index. This can be useful for contributors and integrators to track during a guardian set upgrade. Note that currently Solana and Pythnet are not supported since the dashboard does not have an allocated Solana or Pythnet RPC to use (every other chain has public endpoint usable from a web page). A future PR may move this functionality into a cloud function so that this info can be provided and cached.

### DIFF
--- a/dashboard/src/components/Contracts.tsx
+++ b/dashboard/src/components/Contracts.tsx
@@ -1,0 +1,104 @@
+import {
+  Card,
+  Table,
+  TableBody,
+  TableCell,
+  TableContainer,
+  TableHead,
+  TableRow,
+} from '@mui/material';
+import {
+  Chain,
+  chainToChainId,
+  chainToPlatform,
+  chains,
+  contracts,
+  rpc,
+} from '@wormhole-foundation/sdk-base';
+import { useEffect, useState } from 'react';
+import { useNetworkContext } from '../contexts/NetworkContext';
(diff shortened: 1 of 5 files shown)


panoel (commented):
Do you need this for any of your axios calls?
```
headers: { 'Accept-Encoding': 'application/json' },
```

panoel (approved):

dashboard/src/components/Contracts.tsx
```diff
@@ -0,0 +1,256 @@
...
+                  `0x${Buffer.from(address.bytes, 'base64').toString('hex')}`
+              )
+              .join(','),
+          ]);
+        } catch (e) {}
+      })();
+    } else if (platform === 'Solana') {
+      (async () => {
+        try {
+          // TODO: test this, move to a cloud function
```
panoel, line 87: So, for now, no guardian set index for Solana or Pythnet?

evan-gray: Correct, it will have to be done from a cloud function 😢

evan-gray:
> Do you need this for any of your axios calls?
> 
> ```
> headers: { 'Accept-Encoding': 'application/json' },
> ```

Apparently not 🤷 maybe we’re past the version of axios with that bug.

PR #293, opened 2024-05-15
watcher: de-dup sui digest results

panoel:
Apparently, suix_queryEvents is returning duplicate digests where it hadn't before.  If there are duplicate digests in the sui_multiGetTransactionBlocks call, it throws with a duplicate input error.


bingyuyap (approved):
LGTM!

PR #294, opened 2024-05-16
ft_watcher: make fast_transfer a submodule for easy sync

bingyuyap:
This PR removes the manually copy pasted code from `example-liquidity-layer` and uses git submodules to keep the code up to sync.

1. Use Github submodules to point to `solana-devnet` tagged branch on the `example-liquidity-layer`.
2. Update Github actions to build tarball required for Fast Transfer sdk dependencies. This is important for Github CI actions to run smoothly.
3. Minor `package.json` and `tsconfig.json` fixes.

Please note that ~14,000 lines added and ~27,000 lines deleted were from the `package-lock.json` update. A lot of lines deleted from the previously copied code. 

> NOTE
> 
> Adding steps to build tarball increases the length of the Github CI workflow. In the same workflow, tests fail often (due to RPC errors or whatsoever reasons). The lengthy process will deteriorate developer experience when submitting PR on this repo. We could explore commitment the sdk tarball to the repo to prevent this but it might bloat the repository.


cloud_functions/package.json
```diff
@@ -25,6 +25,7 @@
     "axios": "^1.5.0",
     "borsh": "^1.0.0",
     "dotenv": "^16.0.3",
+    "ethers": "6.12.0",
```
panoel, line 28: I'm not seeing why you needed to add the ethers package?

.github/workflows/build.yaml
```diff
@@ -9,10 +9,21 @@ jobs:
     steps:
       - name: Checkout 🛎️
         uses: actions/checkout@v3
+        with:
+          submodules: recursive
       - uses: actions/setup-node@v3
         with:
           node-version: 16
```
panoel, line 16: Could you change the node-version to 20?  If it breaks things, back out the change and we'll do it in a different PR.

cloud_functions/package.json
bingyuyap: `cloud_fucnctions/getSolanaEvents.ts` uses `ethers.ZeroAddress`. Without this line in the `package.json` it seems to not be able to find the const

.github/workflows/build.yaml
bingyuyap: Changed it. Note that this workflow will only trigger when we push to `main`. But it should be fine since `test.yaml` uses v20 and it works fine

panoel (approved):
I didn't look at the idls.  Also, I don't really know the submodule stuff.

watcher/tsconfig.json
```diff
@@ -2,7 +2,13 @@
   "extends": "../tsconfig.base.json",
   "references": [{ "path": "../common" }],
   "compilerOptions": {
-    "outDir": "dist"
+    "outDir": "dist",
+    "baseUrl": ".",
+    "paths": {
+      "@wormhole-foundation/example-liquidity-layer-solana/*": [
+        "./sdk/example-liquidity-layer/node_modules/@wormhole-foundation/example-liquidity-layer-solana/dist/esm/*"
+      ]
```
barnjamin, line 10: what does this do exactly?

barnjamin: Is that not what the package.json deps do already?

cloud_functions/tsconfig.json
```diff
@@ -3,7 +3,10 @@
   "references": [{ "path": "../common" }, { "path": "../database" }],
   "compilerOptions": {
     "outDir": "dist",
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "module": "Node16",
+    "moduleResolution": "node16",
```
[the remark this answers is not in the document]
bingyuyap: Updated to `NodeNext`

watcher/tsconfig.json
bingyuyap: For some reason I have to include this alias to get it working, else I have to import using the whole path for it to find `matchingEngine` and `tokenRouter`

cloud_functions/tsconfig.json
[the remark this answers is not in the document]
panoel: I believe the VMs are running v20.12.1

watcher/tsconfig.json
barnjamin: This is probably also a `module`/`moduleResoltion` thing  😬

barnjamin (approved):

PR #305, opened 2024-05-30
Make fly work for go 1.21 and devnet

evan-gray:


fly/cmd/heartbeats/main.go
```diff
@@ -60,97 +84,149 @@ type obsvRateRow struct {
...
+
+	env, err := common.ParseEnvironment(*envStr)
+	if err != nil || (env != common.UnsafeDevNet && env != common.TestNet && env != common.MainNet) {
+		if *envStr == "" {
+			logger.Fatal("Please specify --env")
+		}
+		logger.Fatal("Invalid value for --env, should be devnet, testnet or mainnet", zap.String("val", *envStr))
+	}
+
+	// Build the set of guardians based on our environment, where the default is mainnet.
```
panoel, line 169: I don't think this comment is a true statement.

fly/cmd/heartbeats/main.go
```diff
@@ -617,3 +723,11 @@ func handleObsv(idx uint) bool {
 func prompt() {
 	fmt.Print("[C]hains, [G]uardians, [M]essage Counts, [O]bsv Rate, [Q]uit: ")
 }
+
+func getGaugeValue(gauge prometheus.Gauge) (float64, error) {
```
panoel, line 727: Is this used anywhere?

panoel (approved):

fly/cmd/heartbeats/main.go
evan-gray: I think the idea was to populate some of the prometheus metrics into the table (like for the dropped messages). I don't totally mind leaving for now.

PR #310, opened 2024-06-05
cloud_functions: convert env var NETWORK to sdk Network

bingyuyap:
Before this env var `NETWORK` has to conform to the sdkv2 format for the NTT related functions. 

Adding this allows us to use full caps `NETWORK` like `MAINNET` and `TESTNET`. This is to standardize the `NETWORK` env var format across all cloud functions.

### DIFF
--- a/cloud_functions/src/computeTotalSupplyAndLocked.ts
+++ b/cloud_functions/src/computeTotalSupplyAndLocked.ts
@@ -1,24 +1,23 @@
 import {
   NTT_MANAGER_CONTRACT,
   NTT_TOKENS,
-  assertEnvironmentVariable,
+  NTTTotalSupplyAndLockedData,
   derivePda,
+  getCustody,
+  getCustodyAmount,
   getEvmTokenDecimals,
   getEvmTotalSupply,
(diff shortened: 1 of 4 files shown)


cloud_functions/src/computeTotalSupplyAndLocked.ts
```diff
@@ -1,10 +1,10 @@
 import {
   NTT_MANAGER_CONTRACT,
   NTT_TOKENS,
-  assertEnvironmentVariable,
   derivePda,
   getEvmTokenDecimals,
   getEvmTotalSupply,
+  getNetwork,
   normalizeToDecimals,
 } from '@wormhole-foundation/wormhole-monitor-common';
```
panoel, line 9: Can you grab lines 12-14 and put them in here, too?  🙏

panoel (approved):

cloud_functions/src/computeTotalSupplyAndLocked.ts
bingyuyap: Updated!

PR #311, opened 2024-06-05
dashboard: update gateway url

evan-gray:
swap P2P for xLabs' wormchain endpoint.

### DIFF
--- a/dashboard/src/utils/consts.ts
+++ b/dashboard/src/utils/consts.ts
@@ -1,10 +1,12 @@
 import { chainIdToChain, chainIds } from '@wormhole-foundation/sdk-base';
 import { chainToIcon } from '@wormhole-foundation/sdk-icons';
 
-export const WORMCHAIN_URL = 'https://tncnt-eu-wormchain-main-01.rpc.p2p.world';
-export const TESTNET_WORMCHAIN_URL = `https://corsproxy.io/?${encodeURIComponent(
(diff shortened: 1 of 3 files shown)


dashboard/src/utils/consts.ts
```diff
@@ -1,10 +1,12 @@
 import { chainIdToChain, chainIds } from '@wormhole-foundation/sdk-base';
 import { chainToIcon } from '@wormhole-foundation/sdk-icons';
 
-export const WORMCHAIN_URL = 'https://tncnt-eu-wormchain-main-01.rpc.p2p.world';
-export const TESTNET_WORMCHAIN_URL = `https://corsproxy.io/?${encodeURIComponent(
-  'https://gateway.testnet.xlabs.xyz'
-)}`;
+export const WORMCHAIN_URL = import.meta.env.DEV
```
panoel, line 4: Does it make sense at some point to change this over to ```WORMCHAIN_URL[network]```?  Or does vite need a different name?

panoel (approved):

dashboard/src/utils/consts.ts
evan-gray: We would still have to do this special toggle to make local development work so I was on the fence. I'm leaning towards leaving it as is for now, though we could provide a different object based on `import.meta.env.DEV`

bingyuyap (approved):

PR #312, opened 2024-06-05
package.json: clean up refs

panoel:
Main goal was to remove @certusone/wormhole-sdk from imported packages.

### DIFF
--- a/cloud_functions/src/_sdk_circleIntegration.ts
+++ b/cloud_functions/src/_sdk_circleIntegration.ts
@@ -1,5 +1,3 @@
-import { BN } from '@project-serum/anchor';
-
 export enum CircleIntegrationPayload {
   DepositWithPayload = 1,
 }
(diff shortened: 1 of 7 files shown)


bingyuyap (dismissed):

bingyuyap (approved):

cloud_functions/src/_sdk_circleIntegration.ts
```diff
@@ -25,7 +23,7 @@ export function parseCircleIntegrationDepositWithPayload(
     throw new Error('not circle integration payload VAA');
   }
   const tokenAddress = payload.subarray(1, 33);
-  const amount = BigInt(new BN(payload.subarray(33, 65)).toString());
+  const amount = BigInt(`0x${payload.subarray(33, 65).toString('hex')}`);
```
evan-gray, line 26: this changed the type of amount from string to BigInt. I don't think that's what you wanted. Add `.toString()` to the end of the line.

panoel: Are you sure?  It looks like the old way has the .toString() on the new BN and not the BigInt.

evan-gray: ah, I am not sure 😅 you are totally right.

PR #316, opened 2024-06-06
common,cloud_function: add getMissThreshold()

panoel:


common/src/consts.ts
```diff
@@ -95,7 +98,14 @@ export const INITIAL_NTT_DEPLOYMENT_BLOCK_BY_NETWORK_AND_CHAIN: {
   ['Devnet']: {},
 };
 
-export const TOKEN_BRIDGE_EMITTERS: { [key in Chain]?: string } = {
+export function getMissThreshold(date: Date, chainId: ChainId): string {
+  const missThresholdInMins = chainId === toChainId("Scroll") ? 120 : MISS_THRESHOLD_IN_MINS_DEFAULT;
+  const missDate = new Date(date);
```
bingyuyap, line 103: since `data` is already type of `Date` do we still need to do `new Date(date)`?

common/src/consts.ts
```diff
@@ -95,7 +98,14 @@ export const INITIAL_NTT_DEPLOYMENT_BLOCK_BY_NETWORK_AND_CHAIN: {
   ['Devnet']: {},
 };
 
-export const TOKEN_BRIDGE_EMITTERS: { [key in Chain]?: string } = {
+export function getMissThreshold(date: Date, chainId: ChainId): string {
```
bingyuyap, line 101: I wonder if it would be easier too have this function to accept types `(number | string | Chain | ChainId)` and then do a conditional conversion to `ChainId` in this function. 

Like this we get a more versatile function and prevent having to write `toChainId(Number(chain))` everywhere.

bingyuyap (approved):

bingyuyap (approved):
Need to format these files
```
[warn] common/src/consts.ts
[warn] dashboard/src/components/Monitor.tsx
[warn] watcher/scripts/solanaMissedMessageAccounts.ts
```

common/src/consts.ts
panoel: No, we do not.

common/src/consts.ts
panoel: Do we *really* want to make life easier for people?

bingyuyap: It's fine with me either way, we can update the function if and when we actually need the versatility

panoel: I made the change...

bingyuyap: Thanks! Let me do a last check

bingyuyap (dismissed):

common/src/consts.ts
```diff
@@ -1,9 +1,19 @@
...
+  toChainId,
+} from '@wormhole-foundation/sdk-base';
 
 export type Mode = 'vaa' | 'ntt';
 
-export const MISS_THRESHOLD_IN_MINS = 40;
+// This is defined here in an effort to keep the number and text in sync.
+// The default value is not exported because the getMissThreshold() function should be used to get the value.
+const MISS_THRESHOLD_IN_MINS_DEFAULT = 40;
 export const MISS_THRESHOLD_LABEL = '40 minutes';
```
evan-gray, line 15: good idea :+1:

common/src/consts.ts
```diff
@@ -95,6 +105,23 @@ export const INITIAL_NTT_DEPLOYMENT_BLOCK_BY_NETWORK_AND_CHAIN: {
   ['Devnet']: {},
 };
 
+export function getMissThreshold(date: Date, chainish: number | string | Chain | ChainId): string {
+  // We would like chainish to really be a ChainId.
+  let chainId: ChainId;
+  if (typeof chainish === 'string' && !Number.isNaN(Number(chainish))) {
+    // e.g. Handle '1'
+    chainId = toChainId(Number(chainish));
```
evan-gray, line 113: I don't think we should do this - if something does not get an sdk version bump and one of these numbers is not in the list of chain ids, this will throw! It would probably be easier to just `Number(chainish)` when it's a string and not care if it's NaN because it will get the default anyway.

dashboard/src/components/Monitor.tsx
```diff
@@ -242,7 +240,9 @@ function ReobserveCodeContent({ misses }: { misses: MissesByChain }) {
         .map(([chain, info]) => {
           const filteredMisses = showAllMisses
             ? info.messages
-            : info.messages.filter((message) => message.timestamp < missThreshold);
+            : info.messages.filter(
+                (message) => message.timestamp < getMissThreshold(now, toChainId(Number(chain)))
```
evan-gray, line 244: Don't use `toChainId` here, as explained above, the whole front-end may crash if/when it throws. Just pass `chain` since you made it handle strings (which is great btw)

dashboard/src/components/Monitor.tsx
```diff
@@ -106,7 +106,7 @@ function BlockDetail({ chain, message }: { chain: string; message: ObservedMessa
       <Typography gutterBottom>
         Block {message.block}{' '}
         <IconButton
-          href={explorerBlock(network, Number(chain) as ChainId, message.block.toString())}
+          href={explorerBlock(network, toChainId(Number(chain)), message.block.toString())}
```
evan-gray, line 109: ditto, probably not safe - might want to revisit `explorerBlock` later

dashboard/src/components/Monitor.tsx
```diff
@@ -78,7 +78,7 @@ function BlockDetail({ chain, message }: { chain: string; message: ObservedMessa
           gutterBottom
         >
           <IconButton
-            href={explorerTx(network, Number(chain) as ChainId, message.txHash)}
+            href={explorerTx(network, toChainId(Number(chain)), message.txHash)}
```
evan-gray, line 81: ditto

dashboard/src/components/Monitor.tsx
```diff
@@ -313,7 +311,7 @@ function Misses({
             <CollapsibleSection
               key={chain}
               defaultExpanded={false}
-              header={`${chainIdToChain.get(Number(chain) as ChainId)} (${chain}) - ${
+              header={`${chainIdToChain.get(toChainId(Number(chain)))} (${chain}) - ${
```
evan-gray, line 314: same, `toChainId` is not safe here

dashboard/src/components/Monitor.tsx
```diff
@@ -403,7 +399,7 @@ function Monitor({ governorInfo }: { governorInfo?: CloudGovernorInfo | null })
                 );
           return filteredMisses.length === 0
             ? counts
-            : { ...counts, [Number(chain) as ChainId]: filteredMisses.length };
+            : { ...counts, [toChainId(Number(chain))]: filteredMisses.length };
```
evan-gray, line 402: same

dashboard/src/components/Monitor.tsx
```diff
@@ -505,14 +501,14 @@ function Monitor({ governorInfo }: { governorInfo?: CloudGovernorInfo | null })
                     header={
                       <div>
                         <Typography variant="h5" sx={{ mb: 0.5 }}>
-                          {chainIdToChain.get(Number(chain) as ChainId)} ({chain})
+                          {chainIdToChain.get(toChainId(Number(chain)))} ({chain})
```
evan-gray, line 504: likely same for this and below

evan-gray (changes requested):
The review comments might come out of order, but essentially everywhere you used `toChainId` is likely unsafe. It would throw, crashing the given code, when a new chain is added but the corresponding sdk has not been bumped yet (or the front-end deployed after rthe backend, etc). I would revert all of those changes and simply use the number of the chain in the `getMissThreshold` code. It should still be find to compare `chainId === toChainId('Scroll')` when `chainId` is a number.

common/src/consts.ts
panoel: I don't feel comfortable with this because if it is a string, it can be '1' or 'Solana'.  That is not handled by toChainId().  But I have a solution that I believe addresses your concerns.

dashboard/src/components/Monitor.tsx
panoel: Changed back.

dashboard/src/components/Monitor.tsx
panoel: Fixed.

dashboard/src/components/Monitor.tsx
panoel: Changed back.

dashboard/src/components/Monitor.tsx
panoel: Changed back

dashboard/src/components/Monitor.tsx
panoel: Changed back

evan-gray (approved):

common/src/consts.ts
evan-gray: Ah, wasn't considering `Solana` case. Looks good now though, thanks!

PR #317, opened 2024-06-07
dashboard: rm metrics

evan-gray:


panoel (commented):
It's true.  ```AnalyticsOutlined``` is indeed not being used.

evan-gray:
> It's true. `AnalyticsOutlined` is indeed not being used.

whoops

panoel (approved):

PR #321, opened 2024-06-11
watcher: add supervisor debugging

panoel:
Add logging to verify the old workers are going away.

### DIFF
--- a/watcher/src/workers/supervisor.ts
+++ b/watcher/src/workers/supervisor.ts
@@ -40,14 +40,15 @@ function spawnWorker(data: WorkerData) {
 }
 
 function monitorWorkers() {
-  setInterval(() => {
+  setInterval(async () => {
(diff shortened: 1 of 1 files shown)


watcher/src/workers/supervisor.ts
```diff
@@ -40,14 +40,15 @@ function spawnWorker(data: WorkerData) {
 }
 
 function monitorWorkers() {
-  setInterval(() => {
+  setInterval(async () => {
```
evan-gray, line 43: can you set interval on an async? I think you might want to do this in a while with a sleep at the end of every loop

evan-gray: like, it would be possible for these to start colliding as written - typically you instead complete one cycle and then await the delay you want

panoel: Fixed

evan-gray (approved):

PR #323, opened 2024-06-12
ft_watcher: parse auction history if auction is closed

bingyuyap:
PR does a few things:
1. try to get auction from auction account 
    - if auction account is closed, grab the `auction_history` which the auction is in
        - if the auction <> auction history mapping is not indexed in the db, index the database 
         - if the mapping is in db, use this pubkey to query the `auction_history` account
2. query `message_protocol` from `AuctionUpdated` event
3. tests
    - query a closed auction 
    - update `executeFastOrderCctp`  txHash
4. add a table to auction history mapping

### DIFF
--- a/watcher/src/watchers/FTSolanaWatcher.ts
+++ b/watcher/src/watchers/FTSolanaWatcher.ts
@@ -1,6 +1,10 @@
 import { Network } from '@wormhole-foundation/sdk-base';
 import { SolanaWatcher } from './SolanaWatcher';
-import { MatchingEngineProgram } from '@wormhole-foundation/example-liquidity-layer-solana/matchingEngine';
+import {
+  AuctionEntry,
+  AuctionInfo,
+  MatchingEngineProgram,
+} from '@wormhole-foundation/example-liquidity-layer-solana/matchingEngine';
 import {
   ConfirmedSignatureInfo,
   Connection,
@@ -12,7 +16,7 @@ import {
 } from '@solana/web3.js';
 import { findFromSignatureAndToSignature } from '../utils/solana';
 import { makeBlockKey } from '../databases/utils';
-import { BorshCoder, Instruction } from '@coral-xyz/anchor';
+import { BorshCoder, EventParser, Instruction } from '@coral-xyz/anchor';
 import { decodeTransferInstruction } from '@solana/spl-token';
 
 import MATCHING_ENGINE_IDL from '../idls/matching_engine.json';
@@ -56,6 +60,7 @@ export class FastTransferSolanaWatcher extends SolanaWatcher {
   readonly MATCHING_ENGINE_PROGRAM_ID: MatchingEngineProgramId;
   readonly USDC_MINT: USDCMintAddress;
   readonly TOKEN_ROUTER_PROGRAM_ID: TokenRouterProgramId;
+  readonly eventParser: EventParser;
 
   constructor(network: Network, isTest: boolean = false) {
     super(network, false);
@@ -76,6 +81,10 @@ export class FastTransferSolanaWatcher extends SolanaWatcher {
     );
     this.getSignaturesLimit = 100;
     this.logger = getLogger(`fast_transfer_solana_${network.toLowerCase()}`);
+    this.eventParser = new EventParser(
+      new PublicKey(this.MATCHING_ENGINE_PROGRAM_ID),
+      this.matchingEngineBorshCoder
+    );
 
     // hacky way to not connect to the db in tests
     // this is to allow ci to run without a db
@@ -226,7 +235,6 @@ export class FastTransferSolanaWatcher extends SolanaWatcher {
         }
       }
     }
-
     return signatures;
   }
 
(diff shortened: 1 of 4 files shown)


watcher/src/watchers/FTSolanaWatcher.ts
```diff
@@ -647,6 +659,178 @@ export class FastTransferSolanaWatcher extends SolanaWatcher {
...
+      if (!auction.info) {
+        throw new Error('Auction info not found');
+      }
+
+      return {
+        vaaHash: Buffer.from(auction.vaaHash).toString('hex'),
+        info: auction.info,
+      };
+    } catch (e) {
+      try {
```
panoel, line 687: Would it be less overhead to not have embedded try...catch blocks?  Maybe do:
fetchAuction()
if not found, fetch from history
if no history, log and return null
return data.

watcher/src/watchers/FTSolanaWatcher.ts
```diff
@@ -647,6 +659,178 @@ export class FastTransferSolanaWatcher extends SolanaWatcher {
...
+        };
+      } catch (e) {
+        this.logger.error('Failed to fetch auction from history:', e);
+      }
+    }
+
+    return null;
+  }
+
+  async fetchEventFromLogs(name: string, logs: string[]) {
```
panoel, line 706: Does this function need to be async?

watcher/src/watchers/FTSolanaWatcher.ts
```diff
@@ -633,10 +645,10 @@ export class FastTransferSolanaWatcher extends SolanaWatcher {
 
   async parseSettleAuctionNoneLocal(
     res: VersionedTransactionResponse,
-    instruction: MessageCompiledInstruction
+    ix: MessageCompiledInstruction
   ) {
     // Slow relay is not done yet, will implement after
-    throw new Error('[parseSettleAuctionNoneLocal] not implemented');
+    throw new Error('[parseSettleAuctionNoneCctp] not implemented');
```
panoel, line 651: Is the non matching name change intentional?

watcher/src/watchers/__tests__/FTSolanaWatcher.test.ts
```diff
@@ -7,7 +7,7 @@ jest.setTimeout(60_000);
 // It is just an entrypoint to test the whole thing with a local postgres database.
 test.skip('getMessagesByBlock', async () => {
```
panoel, line 8: Do you still want this skipped?

watcher/src/watchers/FTSolanaWatcher.ts
evan-gray: agree, this structure of throwing just to catch it yourself is a bit awkward. could just return when you find it, therefore continuing when you don't and log the error and return null at the bottom (if that's the behavior you want)

watcher/src/watchers/FTSolanaWatcher.ts
evan-gray: maybe it's async to match the calling pattern of other functions, but agree it doesn't *need* to be (though it can be) and an explicit return type would be good.

watcher/src/watchers/FTSolanaWatcher.ts
```diff
@@ -647,6 +659,178 @@ export class FastTransferSolanaWatcher extends SolanaWatcher {
...
+      try {
+        const auctionHistory = await this.matchingEngineProgram.fetchAuctionHistory(
+          latestAuctionHistoryIndex
+        );
+        auctionHistories.push(auctionHistory);
+        latestAuctionHistoryIndex++;
+      } catch (error) {
+        // if no more auction history records to fetch or an error occurred, break the loop
+        this.logger.error('No more auction history records to fetch or an error occurred:', error);
+        break;
```
evan-gray, line 758: I would prefer variables in the `while` condition over control statements. e.g. something like 
```ts
let doneOrErrored = false
while (!doneOrErrored) {
    try {
        ...
    } catch (error) {
        ...
        doneOrErrored = true
    }
}
```

watcher/src/watchers/FTSolanaWatcher.ts
```diff
@@ -647,6 +659,178 @@ export class FastTransferSolanaWatcher extends SolanaWatcher {
...
+  // in the database
+  async getDbLatestAuctionHistoryIndex(): Promise<bigint> {
+    // this is to allow ci to run without a db
+    if (!this.pg) {
+      return 0n;
+    }
+
+    try {
+      const result = await this.pg('auction_history_mapping').max('index as maxIndex').first();
+      console.log(result);
```
evan-gray, line 793: should this console log stay here?

watcher/src/watchers/FTSolanaWatcher.ts
bingyuyap: you guys are right. Initially thought an account can be reopened and other data could be stored, that's why I did it this way. (`matchingEngineProgram.fetchAuction` throws an error when the auction account is null/has the wrong type of data. 

Will change to checking if the account info is null first before fetching from history

watcher/src/watchers/FTSolanaWatcher.ts
bingyuyap: Will update!

watcher/src/watchers/FTSolanaWatcher.ts
bingyuyap: No it was an accident. Will update

watcher/src/watchers/FTSolanaWatcher.ts
bingyuyap: Will update

watcher/src/watchers/FTSolanaWatcher.ts
bingyuyap: Will delete this

panoel (approved):

PR #327, opened 2024-06-19
cloud_functions: catch error and continue

panoel:


bingyuyap (dismissed):

bingyuyap (dismissed):

cloud_functions/src/computeMessageCountHistory.ts
```diff
@@ -40,17 +40,21 @@ export async function computeMessageCountHistory(req: any, res: any) {
           skipRow = false;
           continue;
         }
-        const parsed = deserialize('Uint8Array', signedVAA.data.info.bytes[0].value);
-        if (parsed.timestamp === 0) {
-          // e.g. governance VAAs may have timestamp set to 0
-          continue;
+        try {
```
evan-gray, line 43: note: this try/catch is only safe so long as this endpoint is not cached by something like the last sequence of each emitter. also this error may be indicative of a need to regularly update this endpoint every time a new chain is added, likely introduced with sdkv2. that is... quite undesirable

watcher/src/watchers/__tests__/EVMWatcher.test.ts
```diff
@@ -2,6 +2,7 @@ import { expect, test } from '@jest/globals';
 import { INITIAL_DEPLOYMENT_BLOCK_BY_NETWORK_AND_CHAIN } from '@wormhole-foundation/wormhole-monitor-common';
 import { Block, EVMWatcher, LOG_MESSAGE_PUBLISHED_TOPIC } from '../EVMWatcher';
 import { contracts } from '@wormhole-foundation/sdk-base';
+import { max } from 'bn.js';
```
evan-gray, line 5: this is unused.

evan-gray (approved):

watcher/src/watchers/__tests__/EVMWatcher.test.ts
panoel: Removed

cloud_functions/src/computeMessageCountHistory.ts
panoel: I can call isChain() and continue on bad chain instead of doing the try catch block, if that is any better?

evan-gray: I don't understand why we should skip these at all - do it really require anything more than a number?

panoel: Changed back to the old way.

evan-gray (approved):

PR #328, opened 2024-06-21
database: add retries to coingecko

panoel:


bingyuyap (dismissed):
LGTM. Do we want to cap the back off and number of retries?

bingyuyap (approved):
LGTM. Seems like we have capped the max backoff time and throw if we reach 5 minutes, which caps the number of retries as well. Thanks!

PR #335, opened 2024-07-15
cloud_functions: remove refTime check

panoel:
The dashboard and the cloud_function weren't always identifying the same VAAs as missing.  This puts them in sync.


cloud_functions/src/alarmMissingVaas.ts
```diff
@@ -96,14 +96,7 @@ export async function alarmMissingVaas(req: any, res: any) {
             // Check the timestamp and only send messages that are older than MISS_THRESHOLD_IN_MINS
             const msg: ObservedMessage = msgs.messages[i];
             // If there is a reference time for this chain, use it.  Otherwise, use the current time.
```
bingyuyap, line 98: Do we want to update this comment as well? Seems outdated

bingyuyap (dismissed):

cloud_functions/src/alarmMissingVaas.ts
panoel: Good catch.  Fixed.

bingyuyap (approved):

PR #336, opened 2024-07-17
cloud_functions/dashboard: add getGuardianSetInfo cf

panoel:


cloud_functions/package.json
```diff
@@ -19,6 +19,8 @@
     "@google-cloud/pubsub": "^3.4.1",
     "@google-cloud/storage": "^6.8.0",
     "@solana/web3.js": "^1.87.3",
+    "@wormhole-foundation/sdk-base": "^0.8.0",
```
evan-gray, line 22: Leaving this comment for posterity - this was required as the cloud_functions are currently deployed with `pack`ed versions of their siblings and does not include the workspace package lock. In this circumstance, npm was choosing an older sdk as the direct `node_modules` dependency when we wanted the version used by `database` and explicitly defined by the workspace.

cloud_functions/src/getGuardianSetInfo.ts
```diff
@@ -0,0 +1,52 @@
...
+  const firestoreDb = new Firestore({});
+  try {
+    const collectionRef = firestoreDb.collection(firestoreCollection);
+    const snapshot = await collectionRef.get();
+    snapshot.docs.forEach((doc) => {
+      values[doc.id as Chain] = {
+        timestamp: doc.data().timestamp,
+        contract: doc.data().contract,
+        guardianSet: doc.data().guardianSet,
+        guardianSetIndex: BigInt(doc.data().guardianSetIndex),
```
evan-gray, line 45: could we not convert this to BigInt here so it doesn't have to be converted back on line 25?

common/src/consts.ts
```diff
@@ -293,3 +293,14 @@ export const GUARDIAN_SET_4 = [
     name: 'Staking Facilities',
   },
 ];
+
+export type GuardianSetInfo = {
+  timestamp: string;
+  contract: string;
+  guardianSetIndex: bigint;
```
evan-gray, line 300: akin to the previous comment, is it any easier if this is just a string?

dashboard/src/hooks/useGetGuardianSet.ts
```diff
@@ -1,25 +1,30 @@
...
 import axios from 'axios';
 import { useEffect, useState } from 'react';
 import { useNetworkContext } from '../contexts/NetworkContext';
-import { WORMCHAIN_URL } from '../utils/consts';
-import { queryContractSmart } from '../utils/queryContractSmart';
+import { queryContractSmart } from '@wormhole-foundation/wormhole-monitor-common/src/queryContractSmart';
+import { utils } from '@wormhole-foundation/sdk-solana-core';
 
 function useGetGuardianSet(chain: Chain, address: string | undefined) {
+  console.log('Calling useGetGuardianSet with chain:', chain, 'and address:', address);
```
evan-gray, line 14: should we limit this method back to simply fetching the guardian set from eth since that's all it's used for now?

dashboard/tsconfig.json
```diff
@@ -18,5 +18,5 @@
     "noEmit": true,
     "jsx": "react-jsx"
   },
-  "include": ["src"]
+  "include": ["src", "../common/src/queryContractSmart.ts"]
```
evan-gray, line 21: this shouldn't be needed if properly importing from the package.

dashboard/src/components/Contracts.tsx
```diff
@@ -32,8 +40,44 @@ function CoreBridgeInfo({ chain, address }: { chain: Chain; address: string | un
...
+}
+
 function Contracts() {
   const { currentNetwork } = useNetworkContext();
+  const [guardianSetInfoByChain, setGuardianSetInfoByChain] = useState<GuardianSetInfoByChain>({});
+
+  useEffect(() => {
+    async function fetchData() {
+      const data = await getCoreBridgeInfos();
+      setGuardianSetInfoByChain(data);
```
evan-gray, line 75: This is potentially unsafe (memory leak) to set state after a component is destroyed - this can happen when this user navigates away while the request is outstanding. Check the other hooks for examples of how to set and check a `cancelled` variable. 

Also probably worth moving this to a hook like `useGetGuardianSetInfoByChain` in `hooks`.

evan-gray (changes requested):
Overall good, couple small things 🙂

cloud_functions/package.json
panoel: Added comment

dashboard/tsconfig.json
panoel: Removed.

cloud_functions/src/getGuardianSetInfo.ts
panoel: Converted everything to string

dashboard/src/hooks/useGetGuardianSet.ts
panoel: Changed.

dashboard/src/components/Contracts.tsx
panoel: Moved to a hook.

evan-gray (approved):

PR #339, opened 2024-07-26
Bing/ft watcher worker

bingyuyap:
This PR updates the watcher code such that `ft_watcher` can be run

main changes:
- add `ft` to `Mode` 
- add `INITIAL_DEPLOYMENT_BLOCK` for FT and add a reusable type called `NetworkChainBlockMapping`
- add map to map `INITIAL_DEPLOYMENT_BLOCK` to their modes
- add map to map collection name to their modes
- instead of using `isNTT` boolean use `Mode`
- update to some naming of classes and methods 

db updates:
- `fast_vaa_id` and `fast_vaa_hash` to be `UNIQUE` as at some point in time they might be null

### DIFF
--- a/watcher/src/databases/BigtableDatabase.ts
+++ b/watcher/src/databases/BigtableDatabase.ts
@@ -2,6 +2,7 @@ import { Bigtable } from '@google-cloud/bigtable';
 import { PubSub } from '@google-cloud/pubsub';
 import { Chain, chainToChainId } from '@wormhole-foundation/sdk-base';
 import {
+  Mode,
   assertEnvironmentVariable,
   chunkArray,
   sleep,
@@ -38,7 +39,9 @@ export class BigtableDatabase extends Database {
   firestoreDb: FirebaseFirestore.Firestore;
   latestCollectionName: string;
   latestNTTCollectionName: string;
+  latestFTCollectionName: string;
   pubsubSignedVAATopic: string;
+  collectionNameByMode: { [key in Mode]: string };
   pubsub: PubSub;
   constructor() {
     super();
@@ -46,8 +49,15 @@ export class BigtableDatabase extends Database {
     this.signedVAAsTableId = assertEnvironmentVariable('BIGTABLE_SIGNED_VAAS_TABLE_ID');
     this.vaasByTxHashTableId = assertEnvironmentVariable('BIGTABLE_VAAS_BY_TX_HASH_TABLE_ID');
     this.instanceId = assertEnvironmentVariable('BIGTABLE_INSTANCE_ID');
+    // TODO: make these const?
     this.latestCollectionName = assertEnvironmentVariable('FIRESTORE_LATEST_COLLECTION');
     this.latestNTTCollectionName = assertEnvironmentVariable('FIRESTORE_LATEST_NTT_COLLECTION');
+    this.latestFTCollectionName = assertEnvironmentVariable('FIRESTORE_LATEST_FT_COLLECTION');
+    this.collectionNameByMode = {
+      vaa: this.latestCollectionName,
+      ntt: this.latestNTTCollectionName,
+      ft: this.latestFTCollectionName,
+    };
     this.pubsubSignedVAATopic = assertEnvironmentVariable('PUBSUB_SIGNED_VAA_TOPIC');
     try {
       this.bigtable = new Bigtable();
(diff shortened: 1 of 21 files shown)


watcher/src/databases/BigtableDatabase.ts
```diff
@@ -38,16 +39,25 @@ export class BigtableDatabase extends Database {
...
   pubsubSignedVAATopic: string;
+  collectionNameByMode: { [key in Mode]: string };
   pubsub: PubSub;
   constructor() {
     super();
     this.msgTableId = assertEnvironmentVariable('BIGTABLE_TABLE_ID');
     this.signedVAAsTableId = assertEnvironmentVariable('BIGTABLE_SIGNED_VAAS_TABLE_ID');
     this.vaasByTxHashTableId = assertEnvironmentVariable('BIGTABLE_VAAS_BY_TX_HASH_TABLE_ID');
     this.instanceId = assertEnvironmentVariable('BIGTABLE_INSTANCE_ID');
+    // TODO: make these const?
```
panoel, line 52: I'm fine with keeping these env vars.  I think I would rather not make them consts

watcher/src/watchers/Watcher.ts
```diff
@@ -15,13 +16,18 @@ export class Watcher {
...
     this.network = network;
     this.chain = chain;
-    this.isNTT = isNTT;
-    this.logger = isNTT ? getLogger('NTT_' + chain) : getLogger(chain);
+    this.mode = mode;
+
+    // `vaa` -> ''
+    // `ntt` -> 'NTT_'
+    // `ft` -> 'FT_'
+    const loggerPrefix = mode === 'vaa' ? '' : mode.toUpperCase() + '_';
```
panoel, line 29: Does it make sense to give 'vaa' a free pass on prefix?  Should we make everything have a prefix? WDYT?

watcher/src/watchers/FTSolanaWatcher.ts
```diff
@@ -108,7 +108,7 @@ export class FastTransferSolanaWatcher extends SolanaWatcher {
   }
 
   // TODO: Modify this so watcher can actually call this function (Add enum for mode)
```
panoel, line 110: Is now the time to implement this?  Or a different PR?

watcher/src/watchers/FTEVMWatcher.ts
```diff
@@ -71,8 +145,12 @@ export class FTWatcher extends Watcher {
     this.logger.debug(`saving ${fastTransfers.length} fast transfers`);
 
     // Batch insert the fast transfers
-    await this.pg('market_orders').insert(fastTransfers).onConflict('fast_vaa_id').merge();
+    try {
```
panoel, line 148: Is it okay to continue without saving as if everything is fine?

panoel (approved):

watcher/src/databases/BigtableDatabase.ts
evan-gray: They definitely should not be consts, as 1) this code should not leak infrastructure details and 2) should be portable to anyone's gcloud setup.

watcher/src/watchers/FTSolanaWatcher.ts
bingyuyap: This PR does this. Removing this

watcher/src/watchers/FTEVMWatcher.ts
bingyuyap: added logic to split the batch into smaller batches to reduce impact if any insert fails. Also added retry with exponential backoff + logging to tell us which batch in which block range fails

watcher/src/watchers/FTEVMWatcher.ts
```diff
@@ -56,23 +58,128 @@ export class FTWatcher extends Watcher {
...
+      for (let attempt = 1; attempt <= maxRetries; attempt++) {
+        try {
+          await this.pg('market_orders').insert(batch).onConflict('fast_vaa_id').merge();
+          this.logger.info(
+            `Successfully saved batch ${batchNumber}/${totalBatches} (${batch.length} transfers)`
+          );
+          break;
+        } catch (e) {
+          if (attempt === maxRetries) {
+            this.logger.error(
```
panoel, line 167: Is saving of fast transfers tied to anything?  Like do we need to stop progressing until saves work?  What are the consequences of continuing without saving?

panoel (approved):

watcher/src/watchers/FTEVMWatcher.ts
bingyuyap: saving a fast transfer in evm watcher allows you to know when a user starts the fast transfer. failing to save here means empty data on initiating the fast transfer but the rest of the lifecycle will still be saved. 

it is similar to that of ntt watcher

PR #340, opened 2024-07-26
Deploy Cloud Functions from Workspace

evan-gray:
The cloud functions deployment previously did not include a `package-lock.json` and included a pre-`pack`ed version of `common` and `database`. This lead to some odd dependency issues which were only encountered in the cloud.

This PR changes the deploy to use the top level workspace package. The downside is that all workspace dependencies are now installed, but the upside is that the dependencies are locked and cached in cloud build which, bonus, cuts the cloud build time nearly in half. Local testing environments should now exactly match cloud deployments!

It also fixes a start-up performance degredation where each cloud function was loading all of the code and dependencies of every cloud function. A new `FUNCTION` env var was added so that only the relevant function code would be loaded. It is possible that some introspection could be used, but this seemed relatively explicit and safe.

Developers can now run `FUNCTION=getSolanaEvents npm run cf:dev` from the root for a local functions-framework server and `npm run deploy` to run the deployment script.


cloud_functions/scripts/deploy.sh
```diff
@@ -226,17 +226,17 @@ if [ -z "$SOLANA_RPC" ]; then
...
-gcloud functions --project "$GCP_PROJECT" deploy wormchain-monitor --entry-point wormchainMonitor --gen2 --runtime nodejs18 --trigger-http --no-allow-unauthenticated --timeout 300 --memory 256MB --region europe-west3 --set-env-vars WORMCHAIN_SLACK_CHANNEL_ID=$WORMCHAIN_SLACK_CHANNEL_ID,WORMCHAIN_SLACK_POST_URL=$WORMCHAIN_SLACK_POST_URL,WORMCHAIN_SLACK_BOT_TOKEN=$WORMCHAIN_SLACK_BOT_TOKEN,WORMCHAIN_PAGERDUTY_ROUTING_KEY=$WORMCHAIN_PAGERDUTY_ROUTING_KEY,WORMCHAIN_PAGERDUTY_URL=$WORMCHAIN_PAGERDUTY_URL,NETWORK=$NETWORK
+gcloud functions --project "$GCP_PROJECT" deploy compute-guardian-set-info --entry-point computeGuardianSetInfo --gen2 --runtime nodejs18 --trigger-http --allow-unauthenticated --timeout 300 --memory 256MB --region europe-west3 --set-env-vars NETWORK=$NETWORK,FIRESTORE_GUARDIAN_SET_INFO_COLLECTION=$FIRESTORE_GUARDIAN_SET_INFO_COLLECTION,FUNCTION=computeGuardianSetInfo
```
panoel, line 229: The memory needs to be 512MB

evan-gray: Hadn't changed it, but set it to that now.

panoel (approved):

PR #342, opened 2024-07-31
common: fix queryContractSmart parsing

evan-gray:
This fixes an error I was seeing in the mainnet dashboard from the prefix bytes being the equivalent to the ascii `{`. This adopts the proper solution of using an official library's value parsing while still maintaining a dependency-free implementation.

### DIFF
--- a/common/src/queryContractSmart.ts
+++ b/common/src/queryContractSmart.ts
@@ -32,6 +32,195 @@ export function uint32ToString(val: number): string {
   return Buffer.from(buf).toString('hex');
 }
 
+// https://github.com/confio/cosmjs-types/blob/66e52711914fccd2a9d1a03e392d3628fdf499e2/src/binary.ts#L60-L68
+enum WireType {
+  Varint = 0,
+
+  Fixed64 = 1,
+
+  Bytes = 2,
+
+  Fixed32 = 5,
+}
+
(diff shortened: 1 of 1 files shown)


common/src/queryContractSmart.ts
```diff
@@ -65,6 +65,6 @@ export async function queryContractSmart(
     'ascii'
   );
   return JSON.parse(
-    asciiResponse.substring(asciiResponse.indexOf('{') || asciiResponse.indexOf('['))
+    asciiResponse.substring(asciiResponse.indexOf('{"') || asciiResponse.indexOf('['))
```
panoel, line 68: Is this a recent breakage?  Does this handle all the cases that it needs to?

panoel (dismissed):

common/src/queryContractSmart.ts
evan-gray: I noticed it this morning. This at least fixes the active error, but like I wrote in the description, I have no guarantees it handles all possible cases. I simply expect it to handle one more case than before 😅

evan-gray: Updated to use the real parsing 👍

panoel: Sweet!

panoel (approved):

PR #343, opened 2024-07-31
cloud_functions: improve computeGuardianSetInfo

evan-gray:
This PR improves computeGuardianSetInfo so that it only takes as long as the slowest RPC call. This should be safe as the requests are all going to different RPCs.


bingyuyap (approved):
love me some async

PR #344, opened 2024-07-31
watcher: mock axios for flaky rpcs

evan-gray:
This PR introduces axios mocking in jest tests for flaky RPCs. Simply drop a file in the `mock` path matching the URL, and that data will be returned. Otherwise, the original axios method will be used.


bingyuyap (approved):
it's beautiful

image

PR #345, opened 2024-08-01
fly: add network env var

panoel:


fly/cmd/fly/main.go
```diff
@@ -140,8 +142,16 @@ func verifyEnvVar(key string) string {
...
-	p2pBootstrap = p2p.MainnetBootstrapPeers
+	if network == "mainnet" {
+		p2pNetworkID = p2p.MainnetNetworkId
+		p2pBootstrap = p2p.MainnetBootstrapPeers
+	} else if network == "testnet" {
+		p2pNetworkID = p2p.TestnetNetworkId
+		p2pBootstrap = p2p.TestnetBootstrapPeers
+	} else {
+		p2pNetworkID = p2p.DevnetNetworkId
+		p2pBootstrap = ""
```
bruce-riley, line 153: Should you fatal error here, since it's not going to work without the bootstrap?

fly/cmd/fly/main.go
```diff
@@ -140,8 +142,16 @@ func verifyEnvVar(key string) string {
 
 func main() {
 	loadEnvVars()
-	p2pNetworkID = p2p.MainnetNetworkId
-	p2pBootstrap = p2p.MainnetBootstrapPeers
+	if network == "mainnet" {
```
evan-gray, line 145: Can this be consistent with other services and do something like

```suggestion
	env, err := common.ParseEnvironment(network)
	if err != nil || (env != common.TestNet && env != common.MainNet) {
		logger.Fatal("Invalid value for NETWORK, should be testnet or mainnet", zap.String("val", network))
	}
	p2pNetworkID = p2p.GetNetworkId(env)
	p2pBootstrap, err = p2p.GetBootstrapPeers(env)
	if err != nil {
		logger.Fatal("failed to determine the bootstrap peers from the environment", zap.String("env", string(env)), zap.Error(err))
	}
```

(this example moves it after the logger setup)

panoel: Changed

fly/cmd/fly/main.go
panoel: Changed

evan-gray (approved):

bruce-riley (approved):

PR #349, opened 2024-08-06
Bing/ft swap monitor

bingyuyap:
This PR adds swap layer parsing for Redeem transactions which allows us to:
1. identify the end of a fast transfer
2. identify the output amount, output token and relaying fee

The ID of a `redeem_swap` is `fill_vaa_id` which is generated in the `execute_fast_transfer_*` stages. Another PR will be opened to add the `fill_vaa_id` in the these stages.

Note: that the swap layer is not deployed on `devnet` (as we need mainnet quotes from dexes we will test on mainnet) hence the need for mocked data in the test and lack thereof test for getting swap transactions within range. These will be added once the mainnet deployment is done.


watcher/src/fastTransfer/consts.ts
```diff
@@ -16,6 +16,8 @@ export interface SolanaContractAddresses {
 export interface EthereumContractAddresses {
   TokenRouter: string;
   CircleBridge?: string;
+  // Devnet has no swap layer as they need the mainnet qutoes from Uniswap
```
panoel, line 19: Nit: spelling "quotes"

bingyuyap: fixed!

watcher/src/fastTransfer/types.ts
```diff
@@ -1,6 +1,6 @@
 import { PublicKey } from '@solana/web3.js';
 import BN from 'bn.js'; // Imported since FT codebase uses BN
-
+import { ChainId } from '@wormhole-foundation/sdk-base';
```
panoel, line 3: Is this being used?

watcher/src/watchers/FTEVMWatcher.ts
```diff
@@ -124,11 +132,31 @@ export class FTEVMWatcher extends Watcher {
...
 
-    if (results.length) {
-      await this.saveFastTransfers(results, fromBlock, toBlock);
+    const [tokenRouterResults, swapLayerResults] = await Promise.all([
+      tokenRouterPromise,
+      swapLayerPromise,
+    ]);
+    if (tokenRouterResults.results.length) {
+      try {
+        await this.saveFastTransfers(tokenRouterResults.results, fromBlock, toBlock);
```
panoel, line 144: I don't think this throws.  I think the error is caught before saveFastTransers() returns.  Did you want to propogate any error?

watcher/src/fastTransfer/types.ts
bingyuyap: no its not. thanks for catching that

watcher/src/fastTransfer/swapLayer/parser.ts
```diff
@@ -0,0 +1,106 @@
...
+      .find((event) => event && event.name === 'Redeemed');
+
+    if (!swapEvent) return null;
+
+    return {
+      tx_hash: txHash,
+      recipient: swapEvent.args.recipient,
+      output_amount: BigInt(swapEvent.args.outputAmount.toString()),
+      output_token: swapEvent.args.outputToken,
+      timestamp: new Date(blockTime * 1000),
```
panoel, line 64: Is blockTime in the transaction so that you don't have to pass it in?

watcher/src/fastTransfer/swapLayer/parser.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    // Use AbiCoder to decode the raw input data
+    let fillVaaId: string = '';
+    const abiCoder = new ethers.utils.AbiCoder();
+    try {
+      const decodedInput = abiCoder.decode(['bytes', 'tuple(bytes, bytes, bytes)'], inputData);
+
+      const encodedWormholeMessage = decodedInput[1][0];
+      if (encodedWormholeMessage && encodedWormholeMessage.length >= 8) {
+        const vaaBytes = Buffer.from(encodedWormholeMessage.slice(2), 'hex'); // Remove '0x' if present
```
panoel, line 35: "if present"? Isn't it always there or not there?

watcher/src/watchers/FTEVMWatcher.ts
bingyuyap: updated save redeem swaps to use the swap logic and removed the try catch block here

watcher/src/fastTransfer/swapLayer/parser.ts
bingyuyap: you are right. I didn't see that at first. Updated

bingyuyap: backtracking here: i realized the `tx.timestamp` is undefined on localnet rpc. It's optional on TransactionResponse as well. From what I see on devnet there is no `timestamp` field. I think it's safer to leave it as it is for now. 

```typescript
{
  hash: '0x9a2f04c57a18ab8a75a34f766d89543c7f0b218d8d8ac6fb2c49b808a1e54f57',
  type: 2,
  accessList: [],
  blockHash: '0xe5183774046ce7262efec78c495ab12019fb031e279cc5b80bb335f4672c882c',
  blockNumber: 49505593,
  transactionIndex: 2,
  confirmations: 20039653,
  from: '0xB028e6Fe28743e4E0Db2884A6Fc81ff0C6461847',
  gasPrice: BigNumber { _hex: '0xb8c63f00', _isBigNumber: true },
  maxPriorityFeePerGas: BigNumber { _hex: '0xb2d05e00', _isBigNumber: true },
  maxFeePerGas: BigNumber { _hex: '0xbebc2000', _isBigNumber: true },
  gasLimit: BigNumber { _hex: '0x3a1610', _isBigNumber: true },
  to: '0xe0418C44F06B0b0D7D1706E01706316DBB0B210E',
  value: BigNumber { _hex: '0x00', _isBigNumber: true },
  nonce: 63599,
  data: '0x2a2d7e2100000000000000000000000000000000000000000000000000000000009896800000000000000000000000000000000000000000000000000000000000000001b2621b860cfa5d7f324e06771e4ef1d7f12e9dddf30026d282a78e520625b57100000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000001e848000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004deadbeef00000000000000000000000000000000000000000000000000000000',
  r: '0x47dc4e98fc010a86a4ca5dfd6b233be23a6b326e88cc4caf27941c800e85df83',
  s: '0x6fb2668aa22848e2cfcea62abcddad74dae91acaf6768c4ab8160caec47a2361',
  v: 1,
  creates: null,
  chainId: 421614,
  wait: [Function (anonymous)]
}
```

watcher/src/fastTransfer/swapLayer/parser.ts
bingyuyap: You are right. Updating this silly comment

watcher/src/fastTransfer/swapLayer/parser.ts
panoel: Sounds reasonable to me.

panoel (approved):

PR #358, opened 2024-08-20
cloud_functions: add retries to ntt related requests

bingyuyap:
This PR adds retry and logging for NTT related contract calls

### DIFF
--- a/common/src/utils.ts
+++ b/common/src/utils.ts
@@ -214,3 +214,24 @@ export async function sendToPagerDuty(info: PagerDutyInfo): Promise<any> {
   }
   return response.data.data;
 }
+
+// Retry utility function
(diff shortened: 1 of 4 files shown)


common/src/utils.ts
```diff
@@ -214,3 +214,24 @@ export async function sendToPagerDuty(info: PagerDutyInfo): Promise<any> {
   }
   return response.data.data;
 }
+
+// Retry utility function
+export async function retry<T>(
+  fn: () => Promise<T>,
+  retries = 3,
```
panoel, line 221: I think that if we are going to add a retry() to common, it should, also, take the retries and delay as parameters.

panoel (approved):

panoel (approved):

PR #361, opened 2024-08-21
cloud_functions: add isTokenDenylisted call

panoel:


cloud_functions/src/computeTvlTvm.ts
```diff
@@ -115,6 +117,7 @@ async function populateMaps() {
   const metaData: TokenMetaDatum[] = await getTokenMetadata();
   console.log(`Got ${metaData.length} token metadata entries`);
   for (const md of metaData) {
+    if (isTokenDenylisted(toChainId(toChain(md.token_chain)), md.native_address)) continue;
```
evan-gray, line 120: I think `toChain` will throw if this chain is not in the version of the sdk used by this CF - this could break when a token from a new chain is attested before this cloud function is updated - can that be avoided?

panoel: Good catch.  Fixed.

bingyuyap (approved):

evan-gray (approved):

PR #363, opened 2024-08-26
watcher: reduce Klaytn batch size

panoel:
Not sure why this is a thing now.  But Ankr has reduced the allowed batch size for Klaytn.


bingyuyap (approved):
image
society...

PR #370, opened 2024-09-02
hum: fix negative missing observation count

bingyuyap:
This PR fixes negative observation count in the historical uptime monitor. It seems like RockawayX is sending duplicated observations really quickly such that the bigtable checks are not catching them. Additional dedup logic is needed when calculating missing observations. 

This PR also:
- increases obsC buffer 
- add monitor to check on obsC buffer utilization
for further debugging for the dropped observations bug

### DIFF
--- a/fly/cmd/historical_uptime/main.go
+++ b/fly/cmd/historical_uptime/main.go
@@ -264,6 +264,7 @@ func main() {
 
 	// Add channel capacity checks
 	go monitorChannelCapacity(rootCtx, logger, "obsvC", obsvC)
+	go monitorChannelCapacity(rootCtx, logger, "batchObsvC", batchObsvC)
 
 	// Heartbeat updates
 	heartbeatC := make(chan *gossipv1.Heartbeat, 50)
--- a/fly/cmd/historical_uptime/main.go
+++ b/fly/cmd/historical_uptime/main.go
@@ -259,7 +259,7 @@ func main() {
 	defer rootCtxCancel()
 
 	// Inbound observations
-	obsvC := make(chan *node_common.MsgWithTimeStamp[gossipv1.SignedObservation], 20000)
+	obsvC := make(chan *node_common.MsgWithTimeStamp[gossipv1.SignedObservation], 1024)
 	batchObsvC := make(chan *node_common.MsgWithTimeStamp[gossipv1.SignedObservationBatch], 1024)
 
 	// Add channel capacity checks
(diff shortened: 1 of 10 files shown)


fly/cmd/historical_uptime/main.go
```diff
@@ -406,5 +406,32 @@ func main() {
 	<-rootCtx.Done()
 	logger.Info("root context cancelled, exiting...")
 	// TODO: wait for things to shut down gracefully
+}
 
+func monitorChannelCapacity[T any](ctx context.Context, logger *zap.Logger, channelName string, ch <-chan T) {
```
bruce-riley, line 411: Should monitor `batchObsvC` as well.

fly/pkg/historical_uptime/helpers.go
```diff
@@ -57,14 +60,42 @@ func DecrementMissingObservationsCount(logger *zap.Logger, guardianMissingObserv
...
 				continue
 			}
-			guardianMissingObservations[guardianName][chainID]--
+
+			// Check if we've already processed this guardian for this message
+			if processed[string(messageID)] == nil {
+				processed[string(messageID)] = make(map[string]bool)
+			}
+			if processed[string(messageID)][guardianName] {
+				logger.Warn("Duplicate observation", zap.String("messageID", string(messageID)), zap.String("guardian", guardianName))
```
bruce-riley, line 69: Can't this happen normally during reobservations? Should this just be a debug?

fly/pkg/historical_uptime/helpers.go
```diff
@@ -57,14 +60,42 @@ func DecrementMissingObservationsCount(logger *zap.Logger, guardianMissingObserv
...
+				continue
+			}
+
+			// Mark as processed
+			processed[string(messageID)][guardianName] = true
```
panoel, line 74: Which is less code, using the map[string]bool or using a set of strings instead?  You really aren't using the true/false value at all that I can tell.

panoel (dismissed):

fly/pkg/historical_uptime/helpers.go
```diff
@@ -44,6 +44,9 @@ func InitializeMissingObservationsCount(logger *zap.Logger, messages []*types.Me
 }
 
 func DecrementMissingObservationsCount(logger *zap.Logger, guardianMissingObservations map[string]map[string]int, messageObservations map[types.MessageID][]*types.Observation) {
+	// Keep track of processed observations to avoid duplicates
+	processed := make(map[string]map[string]bool)
```
bruce-riley, line 48: The standard way to do a set in go is `map[string]map[string]struct{}`, then to insert do `processed[key] = struct{}{}`.

fly/pkg/historical_uptime/helpers.go
bingyuyap: No, in a usual case, reobservation that happens after the first observation will just not be persisted as it is marked as duplicated. In our case, there are weird instances where duplicated observations happen at the same time and gets persisted in the database due to race conditions.

I think it should still be a warn as it means something is wrong with our db logic. This doesn't happen in PR #371 anymore but good to have it in for future warnings

fly/pkg/historical_uptime/helpers.go
bingyuyap: using `map[string]map[string]struct{}` as bruce recommended

bruce-riley (approved):

bruce-riley (approved):

PR #371, opened 2024-09-03
hum: batch process obs

bingyuyap:
This PR batches observation for processing to reduce DB calls to bigtable, which is the main bottle neck of the performance for processing. This causes the buffer to max out utilisation and drop observations before we can even process them. 

High level:
1. whenever there is an observation, add to a batch.
2. whenever batch size reaches `batchLimit` (for now its 100) or every 5 seconds, process the batch

High level of batch processing:
1. If message does not exist in cache, grab from db and put it in cache. If not found in db assume we haven't seen it before and create a new one in cache.
2. Grab all observations related and put in cache. Check current observation against in-cache observations for deduplication. If not exist, add to cache
3. After every batch, flush these data to db and clear cache for next batch.

Currently on logs, without batching, a buffer size of 20,000 is at 99-100%. Which causes buffer to drop observations. With this change, the utilisation of the buffer is at 0-2% with a size of 1024.

### DIFF
--- a/fly/cmd/historical_uptime/main.go
+++ b/fly/cmd/historical_uptime/main.go
@@ -5,8 +5,11 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"os/signal"
 	"strconv"
 	"strings"
+	"sync"
+	"syscall"
 	"time"
 
 	node_common "github.com/certusone/wormhole/node/pkg/common"
@@ -287,6 +290,7 @@ func main() {
 	if err != nil {
 		logger.Fatal("Failed to create bigtable db", zap.Error(err))
 	}
+
 	promErrC := make(chan error)
 	// Start Prometheus scraper
 	initPromScraper(promRemoteURL, logger, promErrC)
@@ -306,15 +310,35 @@ func main() {
 	batchSize := 100
 	observationBatch := make([]*types.Observation, 0, batchSize)
 
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
+
+	// to make sure that we wait til observation cleanup is done
+	var wg sync.WaitGroup
+
+	// rootCtx might not cancel if shutdown abruptly
 	go func() {
+		<-sigChan
+		logger.Info("Received signal, initiating shutdown")
+		rootCtxCancel()
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
 		ticker := time.NewTicker(5 * time.Second)
+		defer ticker.Stop()
+
 		for {
 			select {
 			case <-rootCtx.Done():
+				if len(observationBatch) > 0 {
+					historical_uptime.ProcessObservationBatch(*db, logger, observationBatch)
+				}
+				logger.Info("Observation cleanup completed.")
 				return
-			case o := <-obsvC:
+			case o := <-obsvC: // TODO: Rip out this code once we cut over to batching.
 				obs := historical_uptime.CreateNewObservation(o.Msg.MessageId, o.Msg.Addr, o.Timestamp, o.Msg.Addr)
-
 				observationBatch = append(observationBatch, obs)
 
 				// if it reaches batchSize then process this batch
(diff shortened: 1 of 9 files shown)


fly/cmd/historical_uptime/main.go
```diff
@@ -303,23 +303,39 @@ func main() {
...
 			select {
 			case <-rootCtx.Done():
 				return
-			case o := <-obsvC: // TODO: Rip out this code once we cut over to batching.
-				obs := &gossipv1.Observation{
-					Hash:      o.Msg.Hash,
-					Signature: o.Msg.Signature,
-					TxHash:    o.Msg.TxHash,
-					MessageId: o.Msg.MessageId,
+			case o := <-obsvC:
```
bruce-riley, line 315: Just wondering why you ripped out the "TODO"? That is still relevant.

fly/cmd/historical_uptime/main.go
```diff
@@ -303,23 +303,39 @@ func main() {
...
+				// if it reaches batchSize then process this batch
+				if len(observationBatch) >= batchSize {
+					historical_uptime.ProcessObservationBatch(*db, logger, observationBatch)
+					observationBatch = observationBatch[:0] // Clear the batch
 				}
-				historical_uptime.ProcessObservation(*db, logger, o.Timestamp, o.Msg.Addr, obs)
 			case batch := <-batchObsvC:
-				for _, o := range batch.Msg.Observations {
-					historical_uptime.ProcessObservation(*db, logger, batch.Timestamp, batch.Msg.Addr, o)
+				// process immediately since batches are in group
```
bruce-riley, line 326: Not sure this makes sense, since many batches will be small. I would process batches the same as single observations.

fly/cmd/historical_uptime/main.go
```diff
@@ -303,23 +303,39 @@ func main() {
...
 
-	// WIP(bing): add metrics for guardian observations
+	batchSize := 100
+	observationBatch := make([]*types.Observation, 0, batchSize)
+
 	go func() {
+		ticker := time.NewTicker(5 * time.Second)
 		for {
 			select {
 			case <-rootCtx.Done():
```
bruce-riley, line 313: Should you process the batch before you exit?

bruce-riley (commented):
Not quite done yet, but here's what I've got so far.

fly/cmd/historical_uptime/main.go
bingyuyap: that is good info, i'll change it

fly/cmd/historical_uptime/main.go
bingyuyap: you are right, adding that

fly/pkg/historical_uptime/process_observation.go
```diff
@@ -3,53 +3,119 @@ package historical_uptime
...
-func createNewObservation(timestamp time.Time, addr []byte, o *gossipv1.Observation) types.Observation {
-	ga := eth_common.BytesToAddress(addr).String()
-	return types.Observation{
-		MessageID:    types.MessageID(o.MessageId),
-		GuardianAddr: ga,
-		Signature:    hex.EncodeToString(o.Signature),
-		ObservedAt:   timestamp,
-		Status:       types.OnTime,
-	}
+var cache = &bigtable.ObservationCache{
```
bruce-riley, line 16: Could do a `NewObservationCache` up in `cache.go`.

fly/pkg/historical_uptime/process_observation.go
```diff
@@ -3,53 +3,119 @@ package historical_uptime
...
+
+	// Clear the cache after flushing
+	cache.Messages = make(map[types.MessageID]*types.Message)
+	cache.Observations = make(map[types.MessageID]map[string]*types.Observation)
+
+	return nil
+}
+
+// ProcessObservation processes a single observation, updating the cache and checking observation times.
+func ProcessObservation(db bigtable.BigtableDB, logger *zap.Logger, o *types.Observation) {
```
bruce-riley, line 47: Could this be something like `ProcessObservationAlreadyLocked` and you grab the lock once before the for loop at line 23?

PR #374, opened 2024-09-10
hum: add safe and finalized heights

bingyuyap:
This PR adds the `safe` and `finalized` height to prometheus metrics


bruce-riley (dismissed):

bruce-riley (approved):
Thanks for making that change!

PR #377, opened 2024-09-13
cloud_functions: add destChain param to getSolanaEvents

kev1n-peters:


kev1n-peters:
DefiLlama was asking for this

cloud_functions/src/getSolanaEvents.ts
```diff
@@ -12,6 +12,19 @@ import * as ethers from 'ethers';
...
 import { assertEnvironmentVariable, EventData } from '@wormhole-foundation/wormhole-monitor-common';
+import { Chain, ChainId, chainToChainId, isChain } from '@wormhole-foundation/sdk-base';
+
+function convertDefiLlamaChainToChainId(chain: string): ChainId {
+  if (chain === 'avax') {
+    return chainToChainId('Avalanche');
+  }
+  // DefiLlama uses lowercase chain names
+  chain = chain.charAt(0).toUpperCase() + chain.slice(1);
+  if (!isChain(chain as Chain)) {
```
bingyuyap, line 23: I think there's no need for `as Chain` since `isChain` takes string as the argument

```
export declare const isChain: (chain: string) => chain is "Solana" | "Ethereum" | ... | "PolygonSepolia";
```

cloud_functions/src/getSolanaEvents.ts
```diff
@@ -12,6 +12,19 @@ import * as ethers from 'ethers';
...
+    throw new Error('invalid chain');
+  }
+  return chainToChainId(chain as Chain);
```
bingyuyap, line 26: `toChainId(chain)` here would be type-safer. It type guards `Chain` before converting to `ChainId`

```ts
export const toChainId = (chain: number | string): ChainId => {
  switch (typeof chain) {
    case "string":
      if (isChain(chain)) return chainToChainId(chain);
      break;
    case "number":
      if (isChainId(chain)) return chain;
      break;
  }
  throw Error(`Cannot convert to ChainId: ${chain}`);
};
```

kev1n-peters: `isChain` guards chain as `Chain` so I think now with the change you suggested above we're good

bingyuyap (approved):

PR #378, opened 2024-09-20
fix missing message index mutation 

bingyuyap:
This PR does a few things:
- fixes missing observation bug where monitor cannot find unprocessed messages as message indexes weren't created in `FlushCache`. 
- logs the bulk mutation for more observability.
- updates `MAX_CHAIN_ID` to be the largest mainnet chain id.

### DIFF
--- a/fly/cmd/historical_uptime/main.go
+++ b/fly/cmd/historical_uptime/main.go
@@ -109,7 +109,9 @@ var (
 	)
 )
 
-const PYTHNET_CHAIN_ID = int(vaa.ChainIDPythNet)
+// [0, MAX_CHAIN_ID] is the range of chain id that we will track for the uptime monitor
+// in this case it's snaxchain since it's the largest mainnet chain idj
+const MAX_CHAIN_ID = vaa.ChainIDSnaxchain
 
 // guardianChainHeights indexes current chain height by chain id and guardian name
 var guardianChainHeights = make(common.GuardianChainHeights)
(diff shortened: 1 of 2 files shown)


lrogana (approved):

fly/cmd/historical_uptime/main.go
```diff
@@ -187,10 +189,7 @@ func initPromScraper(promRemoteURL string, logger *zap.Logger, errC chan error)
 				case <-t.C:
 					recordGuardianHeightDifferences()
 
-					for i := 1; i < 36; i++ {
-						if i == PYTHNET_CHAIN_ID {
-							continue
-						}
+					for i := 1; i <= int(MAX_CHAIN_ID); i++ {
```
panoel, line 192: You are tracking pythnet, then?

panoel (approved):

fly/cmd/historical_uptime/main.go
bingyuyap: yes, as it seems that the batch processing has enabled us to track pyth without performance degradation

fly/cmd/historical_uptime/main.go
```diff
@@ -109,7 +109,9 @@ var (
 	)
 )
 
-const PYTHNET_CHAIN_ID = int(vaa.ChainIDPythNet)
+// [0, MAX_CHAIN_ID] is the range of chain id that we will track for the uptime monitor
+// in this case it's snaxchain since it's the largest mainnet chain idj
+const MAX_CHAIN_ID = vaa.ChainIDSnaxchain
```
bruce-riley, line 114: This is a maintenance headache. We should think about adding something in the go sdk to give you a max value that will be updated whenever we add a new chain. Or is this meant to match the dashboard / fly? Should we have a shared constant somewhere?

bruce-riley (approved):

fly/cmd/historical_uptime/main.go
bingyuyap: Will address this in a separate PR

PR #381, opened 2024-10-10
Add unichain testnet support

bruce-riley:


common/src/explorer.ts
```diff
@@ -132,6 +132,8 @@ export const explorerBlock = (network: Network, chainId: ChainId, block: string)
     ? `https://bartio.beratrail.io/block/${block}`
     : chainId === chainToChainId('Snaxchain')
     ? `https://explorer-snaxchain-s50q0kjngn.t.conduit.xyz/${block}`
+    : chainId === chainToChainId('Unichain')
+    ? `https://sepolia.unichain.org/${block}`
```
panoel, line 136: This explorer link doesn't work.  Does one even exist?

panoel (changes requested):
dashboard/package.json needs to be updated with a version of @wormhole/sdk-icons that has the new Unichain icon.

common/src/explorer.ts
bruce-riley: Oops. That was the RPC, not the explorer. Fixed it.

panoel (approved):

PR #382, opened 2024-10-24
Add World Chain and Monad Devnet support

bruce-riley:


watcher/src/index.ts
```diff
@@ -55,6 +55,8 @@ const supportedChains: Chain[] =
         'Berachain',
         'Snaxchain',
         'Unichain',
+        'Worldchain',
```
panoel, line 58: Maybe comment out this line until they are supported?

panoel (approved):

watcher/src/index.ts
bruce-riley: After and offline discussion, we decided to leave this in since Worldchain should be deployed to testnet soon.

PR #384, opened 2024-11-11
Ft/fix swap layer

bingyuyap:
This PR attempts to fix everything once the mainnet contracts are fixed:

- `redeem_time` is nuillable since `complete_x_payload` is not considered as a real redeem until `release_inbound` is called (which is linked via `staged_inbound`)
- save `TransactionComplete` batch from Solana Swap Layer parser into db


panoel:
This is obsolete.

PR #396, opened 2025-01-14
Add Monad testnet support

bruce-riley:


database/src/token_bridge/address.ts
```diff
@@ -60,7 +60,11 @@ export const getNativeAddress = async (
       );
     } else if (tokenChain === chainToChainId('Aptos')) {
       const wh = await wormhole('Mainnet', [aptos]);
-      const client = new AptosClient('https://fullnode.mainnet.aptoslabs.com');
+      const config = new AptosConfig({
+        fullnode: 'https://fullnode.mainnet.aptoslabs.com',
+        network: AptosNetwork.MAINNET,
+      });
+      const client = new Aptos();
```
panoel, line 67: I think you need to pass the config into the Aptos constructor.

```suggestion
      const client = new Aptos(config);
```

panoel (changes requested):

panoel (approved):

PR #397, opened 2025-01-14
watcher: FT update IDLs

panoel:
The IDLs were copied from  https://github.com/wormhole-foundation/example-liquidity-layer/blob/main/solana/ts/src/idl/json/

### DIFF
--- a/watcher/src/utils/solana.ts
+++ b/watcher/src/utils/solana.ts
@@ -1,5 +1,13 @@
 import { BN } from '@coral-xyz/anchor';
-import { Connection, SolanaJSONRPCError, VersionedBlockResponse } from '@solana/web3.js';
+import {
+  Connection,
+  PublicKey,
+  SolanaJSONRPCError,
+  VersionedBlockResponse,
+  VersionedMessage,
+  VersionedTransactionResponse,
+} from '@solana/web3.js';
+import { isLegacyMessage } from '@wormhole-foundation/wormhole-monitor-common';
 
 export const findNextValidBlock = async (
   connection: Connection,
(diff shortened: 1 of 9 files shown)


bruce-riley (approved):

watcher/src/watchers/FTSolanaWatcher.ts
```diff
@@ -267,6 +266,7 @@ export class FTSolanaWatcher extends SolanaWatcher {
     instruction: MessageCompiledInstruction,
     seq: number
   ): Promise<void> {
+    const allKeys = await getAllKeys(this.getConnection(), res);
```
evan-gray, line 269: This should be done before searching for the `programIdIndex` (above line 243) in `fetchAndProcessMessages`, as that will fail if the program is specified in the lookup table. You can then pass that through to here.

watcher/src/utils/solana.ts
```diff
@@ -70,3 +78,50 @@ export const U64 = {
   },
   from: (amount: BN, unit: number) => amount.toNumber() / unit,
 };
+
+export async function getAllKeys(
```
evan-gray, line 82: nice. can you switch the Solana watcher to use this function as well?

evan-gray (changes requested):

watcher/src/watchers/FTSolanaWatcher.ts
panoel: Fixed.

watcher/src/utils/solana.ts
panoel: Changed.

evan-gray (approved):

PR #400, opened 2025-01-31
Change ft merge

panoel:
This PR:
- updates how data is merged into the pg db for EVM.
- updates the imported code from liquidity layer.
- uses a different approach to look for fast transfers on EVM.
- picks up a newer version of the example-liquidity-layer repo in the submodule.


watcher/src/watchers/FTEVMWatcher.ts
```diff
@@ -129,84 +140,284 @@ export class FTEVMWatcher extends Watcher {
...
+        `Log tx: ${log.transactionHash}, block: ${log.blockNumber} is not a fast market order`
+      );
+      return null;
+    }
+
+    this.logger.debug(`Log tx: ${txHash}, block: ${log.blockNumber} IS a fast market order`);
+    return this.evmTokenRouter.getTransactionResults(txHash);
+  }
+
+  async fetchBlockTime(
```
evan-gray, line 352: can code be shared with the EVMWatcher?

watcher/src/watchers/FTEVMWatcher.ts
```diff
@@ -129,84 +140,284 @@ export class FTEVMWatcher extends Watcher {
...
+  ): Promise<number> {
+    let block = blocks.get(blockNumber);
+    if (!block) {
+      block = await this.provider.getBlock(blockNumber);
+      blocks.set(blockNumber, block);
+    }
+    return block.timestamp;
+  }
+
+  async getLogs(
```
evan-gray, line 364: ditto

evan-gray (approved):
just gave a high-level pass. could be a couple things worth cleaning up, but could be merged as is given that it's independent.

.github/workflows/build.yaml
```diff
@@ -4,6 +4,48 @@ on:
     branches:
       - main
 jobs:
+  verify-checksums:
```
evan-gray, line 7: I think you wanted to add this to test. Build only runs on pushes to `main` (for building and deploying the dashboard frontend)

.github/workflows/build.yaml
```diff
@@ -4,6 +4,48 @@ on:
...
+
+      - name: Install Foundry (Forge & Cast)
+        run: |
+          curl -L https://foundry.paradigm.xyz | bash
+          export PATH="$HOME/.foundry/bin:$PATH"
+          ~/.foundry/bin/foundryup
+          echo "PATH=$HOME/.foundry/bin:$PATH" >> $GITHUB_ENV  # Persist PATH
+
+      - name: Run Makefile to generate files
+        run: make init-submodule && make build-ft-sdk
```
evan-gray, line 24: I don't think you need to run `make init-submodule` if you change the checkout step to include the submodule with 
```
      - uses: actions/checkout@v4
        with:
          submodules: recursive
```

.github/workflows/build.yaml
```diff
@@ -4,6 +4,48 @@ on:
...
+  verify-checksums:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: sudo apt-get update && sudo apt-get install -y make
+
+      - name: Install Foundry (Forge & Cast)
```
evan-gray, line 16: There's an action for this 
```
      - name: Install Foundry
        uses: foundry-rs/foundry-toolchain@v1
        with:
          version: nightly
```

.github/workflows/build.yaml
```diff
@@ -4,6 +4,48 @@ on:
...
+      - name: Compute checksums of committed files
+        run: |
+          mkdir -p committed-checksums
+          cd ./watcher/sdk
+          sum *definitions-*.tgz *evm-*.tgz *solana-*.tgz > ../../committed-checksums/sums.txt
+          cd -
+
+      - name: Compare checksums
+        run: |
+          if ! diff -u generated-checksums/sums.txt committed-checksums/sums.txt; then
```
evan-gray, line 42: I don't think you need to manually do these checksums. They're committed, so just git diff them, e.g. https://github.com/wormhole-foundation/wormhole/blob/54fc8aabff7ec5302c7d90c5558958c971e04e33/.github/workflows/build.yml#L250
```
          git diff --name-only --exit-code && echo "✅ Generated CLI docs match committed docs" || (echo "❌ Generated CLI docs differs from committed CLI docs, run \`make docs\` and commit the result" >&2 && exit 1)
```

evan-gray (changes requested):

.github/workflows/build.yaml
panoel: Changed

.github/workflows/build.yaml
panoel: Changed.

.github/workflows/build.yaml
panoel: Moved.

watcher/src/watchers/FTEVMWatcher.ts
panoel: This is being addressed in a separate PR.

.github/workflows/build.yaml
evan-gray: oh, I guess they did finally make a stable version 🤔 was just used to when that wasn't a thing I guess.

evan-gray (approved):

PR #402, opened 2025-02-10
watcher: refactor base watcher classes.

panoel:


panoel:
This is handled by PR #407

PR #406, opened 2025-02-20
FT: add more mainnet chains

panoel:


watcher/src/fastTransfer/consts.ts
```diff
@@ -78,11 +98,17 @@ export const FAST_TRANSFER_CONTRACTS: FastTransferContractAddresses = {
...
 
 // Separate testnet and mainnet chains
-export type FTEVMMainnetChain = 'Arbitrum' | 'Base';
+export type FTEVMMainnetChain = 'Arbitrum' | 'Base' | 'Ethereum' | 'Optimism' | 'Polygon';
 export type FTEVMTestnetChain = 'ArbitrumSepolia' | 'OptimismSepolia';
 export type FTEVMChain = FTEVMMainnetChain | FTEVMTestnetChain;
 
-export const FTEVMMainnetChains: FTEVMMainnetChain[] = ['Arbitrum', 'Base'];
+export const FTEVMMainnetChains: FTEVMMainnetChain[] = [
+  'Arbitrum',
```
evan-gray, line 106: why is avalanche in the contracts above but not in this list?

evan-gray (dismissed):

watcher/src/fastTransfer/consts.ts
panoel: Because they had me add it and then decided there wouldn't be any fast transfers on avalanche.  So, I left it in case they change their minds in the future.

evan-gray (approved):
In the future, could derive the types of `FTEVMMainnetChain` and `FTEVMTestnetChain` from `FTEVMMainnetChains` and `FTEVMTestnetChains` instead

PR #412, opened 2025-03-11
PROT-9: add alarmFastTransfer cloud function

panoel:


evan-gray (approved):
didn't look super meticulously at the query, but the flow seems reasonable and I assume this will be tested.

PR #416, opened 2025-03-14
cloud_functions: getQuorumHeight

evan-gray:
This new function gets the quorum height for latest, safe, and finalized per the guardian heartbeats cached in firestore.

Deployed to [testnet](https://europe-west3-wormhole-message-db-testnet.cloudfunctions.net/get-quorum-height?chainId=24) and [mainnet](https://europe-west3-wormhole-message-db-mainnet.cloudfunctions.net/get-quorum-height?chainId=24)


cloud_functions/scripts/deploy.sh
```diff
@@ -133,6 +133,7 @@ gcloud functions --project "$GCP_PROJECT" deploy compute-missing-vaas --entry-po
 gcloud functions --project "$GCP_PROJECT" deploy compute-ntt-rate-limits --entry-point computeNTTRateLimits --gen2 --runtime nodejs22 --trigger-http --allow-unauthenticated --timeout 300 --memory 512MB --region europe-west3 --set-env-vars NETWORK=$NETWORK,FUNCTION=computeNTTRateLimits
 gcloud functions --project "$GCP_PROJECT" deploy compute-total-supply-and-locked --entry-point computeTotalSupplyAndLocked --gen2 --runtime nodejs22 --trigger-http --allow-unauthenticated --timeout 300 --memory 512MB --region europe-west3 --set-env-vars NETWORK=$NETWORK,FUNCTION=computeTotalSupplyAndLocked
 gcloud functions --project "$GCP_PROJECT" deploy get-ntt-rate-limits --entry-point getNTTRateLimits --gen2 --runtime nodejs22 --trigger-http --allow-unauthenticated --timeout 300 --memory 512MB --region europe-west3 --set-env-vars NETWORK=$NETWORK,FUNCTION=getNTTRateLimits
+gcloud functions --project "$GCP_PROJECT" deploy get-quorum-height --entry-point getQuorumHeight --gen2 --runtime nodejs22 --trigger-http --allow-unauthenticated --timeout 300 --memory 512MB --region europe-west3 --set-env-vars NETWORK=$NETWORK,FUNCTION=getQuorumHeight
```
panoel, line 136: CLOUD_FUNCTIONS_REFRESH_TIME_INTERVAL should be on this line, too.

cloud_functions/src/getQuorumHeight.ts
```diff
@@ -0,0 +1,122 @@
...
+interface HeartbeatNetwork {
+  id: number;
+  height: string;
+  contractAddress: string;
+  errorCount: string;
+  safeHeight: string;
+  finalizedHeight: string;
+}
+
+const isTestnet = process.env.NETWORK === 'TESTNET';
```
panoel, line 26: I've been using assertEnvironmentVariable() to catch cases where the deploy script is missing something.  Same comment for other env vars

cloud_functions/src/getQuorumHeight.ts
```diff
@@ -0,0 +1,122 @@
...
+  } catch (e) {
+    console.error(e);
+  }
+  return heartbeats;
+}
+
+let cache = { heartbeats: [] as Heartbeat[], lastUpdated: Date.now() };
+// default refresh interval = 15 sec
+const REFRESH_TIME_INTERVAL =
+  Number(process.env.CLOUD_FUNCTIONS_REFRESH_TIME_INTERVAL) || 1000 * 15;
```
panoel, line 50: This is optional?  Currently, we would have to re-deploy to change it or pick up the env var.  I suppose this is an argument for all cloud functions to use the local env in gcp instead of always specifying them in the deploy script?

panoel (dismissed):

cloud_functions/src/getQuorumHeight.ts
evan-gray: I copied this from `getLatestBlocks` - I think what is in the deployment script is what would get set into gcp for that function (i.e. the "local env" *is* what is set in the deploy script). https://cloud.google.com/sdk/gcloud/reference/functions/deploy

cloud_functions/src/getQuorumHeight.ts
evan-gray: fixed, thanks

cloud_functions/scripts/deploy.sh
evan-gray: I was torn on this because I'm not sure they should all be the same, but I added it for now.

cloud_functions/src/getQuorumHeight.ts
```diff
@@ -0,0 +1,125 @@
...
+          const network = heartbeat.networks.find((n) => n.id === chainId);
+          latestHeights.push(BigInt(network?.height || '0'));
+          safeHeights.push(BigInt(network?.safeHeight || '0'));
+          finalizedHeights.push(BigInt(network?.finalizedHeight || '0'));
+        }
+      }
+      latestHeights.sort((a, b) => (a > b ? -1 : a < b ? 1 : 0));
+      safeHeights.sort((a, b) => (a > b ? -1 : a < b ? 1 : 0));
+      finalizedHeights.sort((a, b) => (a > b ? -1 : a < b ? 1 : 0));
+      const quorumIdx = getQuorumCount(latestHeights.length) - 1;
```
am-os, line 109: Is this polling the 13th highest height? Should we remove the `- 1` at the end

PR #417, opened 2025-03-17
Add Mezo testnet support

bruce-riley:


watcher/src/watchers/Watcher.ts
```diff
@@ -43,7 +43,8 @@ export class Watcher {
       chain === 'Ink' ||
       chain === 'HyperEVM' ||
       chain === 'Seievm' ||
-      chain === 'Xlayer'
+      chain === 'Xlayer' ||
+      chain === 'Mezo'
```
panoel, line 47: This probably doesn't need to be 10.  The default of 100 is probably fine.  This kind of thing can only be found out by letting it run.

panoel (approved):

panoel (approved):

PR #421, opened 2025-03-21
watcher: solana shim support

panoel:


watcher/src/watchers/SolanaWatcher.ts
```diff
@@ -162,39 +184,75 @@ export class SolanaWatcher extends Watcher {
...
         for (const instruction of whInstructions) {
-          // skip if not postMessage instruction
           const instructionId = instruction.data;
-          if (instructionId[0] !== 0x08 && instructionId[0] !== 0x01) continue;
+          if (
+            instruction.programIdIndex === programIdIndex &&
+            instructionId[0] === POST_MESSAGE_UNRELIABLE_INSTRUCTION_ID
+          ) {
+            // TODO:  Do I need to verify that this message has no data?
+            // This is an unreliable wormhole message.  It is only used in conjunction with a shim message.
```
evan-gray, line 219: this comment is not inherently true, an unreliable message is used by the shim, but others can use it. perhaps we mean that we only care about the one which is used by the shim? that would be fair for monitoring purposes

watcher/src/watchers/SolanaWatcher.ts
```diff
@@ -162,39 +184,75 @@ export class SolanaWatcher extends Watcher {
...
+            const acctInfo = await this.getConnection().getAccountInfo(accountId, COMMITMENT);
+            if (!acctInfo?.data) throw new Error('No data found in message account');
+            const deserializedMsg = deserializePostMessage(new Uint8Array(acctInfo.data));
+            emitterAddress = deserializedMsg.emitterAddress;
+            sequence = deserializedMsg.sequence;
+          } else {
+            // instruction.programIdIndex === shimProgramIdIndex
+            console.log('Got shim instruction');
+            gotShim = true;
+            const parsedMsg = this.parseShimMessage(instruction.data);
```
evan-gray, line 243: isn't this instruction the core bridge message though, not the shim message (based on the check on line 227)?

evan-gray (changes requested):
I think this can simply look for `POST_MESSAGE_INSTRUCTION_ID` the old way or shim program instructions which start with the shim event discriminator since all we need is the emitter address and sequence.

this appears to be close-ish to what's happening here but not quite. I think it simply needs to be something like

```typescript
if (
  instruction.programIdIndex === programIdIndex &&
  instructionId[0] === POST_MESSAGE_INSTRUCTION_ID
) {
  // core bridge reliable message, do old stuff
} else if (
  instruction.programIdIndex === shimProgramIdIndex &&
  hexData.startsWith(shimMessageEventDiscriminator)
) {
  // shim event, parse data to get the emitter and sequence
}
```

watcher/src/watchers/SolanaWatcher.ts
panoel: Nope.  At 227 we have an unreliable message (that I should make sure is empty to indicate a shim is forthcoming).  Then we continue the for loop on line 219, looking at other instructions.  At 234 we check to see if we have a core and then process that.  Failing that, it must be a shim (because those are the only 2 program IDs we collect).

panoel:
> I think this can simply look for `POST_MESSAGE_INSTRUCTION_ID` the old way or shim program instructions which start with the shim event discriminator since all we need is the emitter address and sequence.
> 
> this appears to be close-ish to what's happening here but not quite. I think it simply needs to be something like
> 
> ```ts
> if (
>   instruction.programIdIndex === programIdIndex &&
>   instructionId[0] === POST_MESSAGE_INSTRUCTION_ID
> ) {
>   // core bridge reliable message, do old stuff
> } else if (
>   instruction.programIdIndex === shimProgramIdIndex &&
>   hexData.startsWith(shimMessageEventDiscriminator)
> ) {
>   // shim event, parse data to get the emitter and sequence
> }
> ```

I think this is exactly what the code is doing.  There are a few extra safety checks, but I see the code doing just that.

watcher/src/watchers/SolanaWatcher.ts
bruce-riley: I think you need to check for the payload length being zero. If it is, then assume this is part of a shim transaction.

panoel: What if I just delete this whole if statement and body?  Line 235 handles this case.

bruce-riley:
Okay, maybe what @evan-gray said earlier today is finally sinking in.

Since the only things the monitor needs is the emitter address and sequence number, you can simply search the inner instructions for something from the shim contract that starts with `e445a52e51cb9a1d441b8f004d4c8970`, which means it's a Shim `MessageEvent`.

Are that and ignoring unreliable messages with no payload are all that is needed for this change??

@panoel sorry I steered you down the path of doing a full-blown implementation like the guardian watcher needs!

watcher/src/watchers/SolanaWatcher.ts
```diff
@@ -18,16 +18,37 @@ import {
...
 const GET_SIGNATURES_LIMIT = 1000;
 
+const ShimContracts: { [key in Network]: string } = {
+  Mainnet: '',
+  Testnet: 'EtZMZM22ViKMo4r5y4Anovs3wKQ2owUmDpjygnMMcdEX',
+  Devnet: '',
+};
+
+const POST_MESSAGE_INSTRUCTION_ID = 0x01;
+const POST_MESSAGE_UNRELIABLE_INSTRUCTION_ID = 0x08;
```
evan-gray, line 37: this is unused now

watcher/src/watchers/SolanaWatcher.ts
```diff
@@ -161,40 +186,71 @@ export class SolanaWatcher extends Watcher {
...
+              i.programIdIndex === shimProgramIdIndex
+          );
 
         const blockKey = makeBlockKey(
           res.slot.toString(),
           new Date(res.blockTime * 1000).toISOString()
         );
 
+        let needShim = false;
+        let gotShim = false;
```
evan-gray, line 217: I don't think `needShim` or `gotShim` are needed

watcher/src/watchers/SolanaWatcher.ts
```diff
@@ -161,40 +186,71 @@ export class SolanaWatcher extends Watcher {
...
+            if (!parsedMsg) {
+              console.log('Failed to parse shim message');
+              continue;
+            }
+            emitterAddress = parsedMsg.emitterAddress;
+            sequence = parsedMsg.sequence;
+          }
+          // TODO: should I check if needShim === gotShim?
 
           vaaKeys.push(
```
evan-gray, line 251: it's possible we reach this code but `emitterAddress` or `sequence` are unset. can we move this into the branches instead of using the control statements or fix the type that they may be undefined and check before pushing?

watcher/src/watchers/SolanaWatcher.ts
```diff
@@ -161,40 +186,71 @@ export class SolanaWatcher extends Watcher {
...
+          let sequence: bigint;
 
-          const acctInfo = await this.getConnection().getAccountInfo(accountId, COMMITMENT);
-          if (!acctInfo?.data) throw new Error('No data found in message account');
-          const { emitterAddress, sequence } = deserializePostMessage(
-            new Uint8Array(acctInfo.data)
-          );
+          if (instruction.programIdIndex === coreBridgeProgramIdIndex) {
+            if (instructionId[0] !== POST_MESSAGE_INSTRUCTION_ID) {
+              console.log('Got non-post message instruction');
```
evan-gray, line 227: probably don't want a console log in the watcher. also I'd avoid the `continue` by just explicitly checking for this in the if and changing the else to an else if

watcher/src/watchers/SolanaWatcher.ts
evan-gray: I may have misread which statement the else was for, still I'd use a more explicit else if

watcher/src/watchers/SolanaWatcher.ts
```diff
@@ -161,40 +186,71 @@ export class SolanaWatcher extends Watcher {
...
+            const accountId = accountKeys[instruction.accountKeyIndexes[1]];
+
+            const acctInfo = await this.getConnection().getAccountInfo(accountId, COMMITMENT);
+            if (!acctInfo?.data) throw new Error('No data found in message account');
+            const deserializedMsg = deserializePostMessage(new Uint8Array(acctInfo.data));
+            emitterAddress = deserializedMsg.emitterAddress;
+            sequence = deserializedMsg.sequence;
+          } else {
+            // instruction.programIdIndex === shimProgramIdIndex
+            console.log('Got shim instruction');
```
evan-gray, line 239: probably don't want the console log here.

watcher/src/watchers/SolanaWatcher.ts
```diff
@@ -217,6 +273,25 @@ export class SolanaWatcher extends Watcher {
...
+  } | null {
+    // First step is to convert the data into a hex string
+    const hexData = encoding.hex.encode(data);
+
+    // Next, we need to look for the discriminator that we care about.
+    if (hexData.startsWith(shimMessageEventDiscriminator)) {
+      // Use the binary layout to deserialize the data
+      const decoded = deserialize(shimMessageEventLayout, data);
+      const emitterAddress = new UniversalAddress(decoded.emitterAddress);
+      const sequence = BigInt(decoded.sequence);
```
evan-gray, line 288: isn't the decoded sequence already a bigint?

evan-gray (changes requested):
maybe a little cleanup and avoid the control statements (which I admit is a personal preference to avoid)

watcher/src/watchers/SolanaWatcher.ts
panoel: Removed.

watcher/src/watchers/SolanaWatcher.ts
panoel: Removed

watcher/src/watchers/SolanaWatcher.ts
panoel: Changed

panoel: Changed control

watcher/src/watchers/SolanaWatcher.ts
panoel: Yes, changed

watcher/src/watchers/SolanaWatcher.ts
panoel: Removed

bruce-riley (approved):

evan-gray (approved):

PR #424, opened 2025-03-28
cloud_functions: add FT alarms

panoel:


cloud_functions/src/alarmFastTransfer.ts
```diff
@@ -117,13 +127,137 @@ async function getDelayedOrders(): Promise<DisplayRow[]> {
...
+    executionTime: row.executionTime,
+    settlementTime: row.settlementTime,
+    auctionAccount: row.auctionAccount,
+    amountIn: BigInt(row.amountIn),
+    amountOut: row.amountOut ? BigInt(row.amountOut) : undefined,
+  }));
+}
+
 type DisplayRow = {
   sourceChain: number; // from fast_vaa_id
```
evan-gray, line 252: FWIW, I think you're technically setting source chain and sequence to strings above but the db probably has those fields as `any`.

evan-gray (approved):

PR #430, opened 2025-06-12
Add Converge support

bruce-riley:


panoel:
Do we want to remove Xpla from `database/src/token_bridge/address.ts::queryExternalId`?
That will allow us to remove the `@xpla/xpla.js` import.

panoel (approved):

PR #436, opened 2025-07-17
add Plume

aapclark:
- update wormhole sdk to 2.4.0 for Plume support

- add Plume mainnet and testnet to common

- add Plume mainnet and testnet to watcher


watcher/src/consts.ts
```diff
@@ -60,6 +60,7 @@ export const RPCS_BY_CHAIN: { [key in Network]: { [key in Chain]?: string } } =
     Worldchain: process.env.WORLDCHAIN_RPC || 'https://worldchain-mainnet.g.alchemy.com/public',
     Ink: process.env.INK_RPC || 'https://rpc-qnd.inkonchain.com',
     HyperEVM: process.env.HYPER_EVM_RPC || 'https://rpc.hyperliquid.xyz/evm',
+    Plume: process.env.PLUME_EVM_RPC || 'https://explorer.plume.org',
```
panoel, line 63: I think this is an explorer and NOT an rpc.  Same comment for the testnet one below.

panoel (changes requested):

watcher/src/consts.ts
aapclark: Fixed.

PR #437, opened 2025-07-17
watcher: upgrade Sui client

panoel:
https://linear.app/wormholelabs/issue/CORE-826/wormhole-dashboard-update-sui-dependency


evan-gray (changes requested):
seems right and the tests pass, but the package-lock change is missing - it likely works because the newer sdk is available in the node modules and the import is different.

evan-gray (approved):

PR #510, opened 2026-05-19
watcher: upgrade sui sdk and use graphql and grpc for the watcher implementation

M-Picco:
Sui JSON RPC is getting deprecated by the end of July, so replace the Sui Watcher json rpc client usage with grpc and graphql. Use default sui providers unless `SUI_GRAPHQL_URL` or `SUI_GRPC_URL` are present on the env vars.

### DIFF
--- a/watcher/src/watchers/__tests__/SuiWatcher.test.ts
+++ b/watcher/src/watchers/__tests__/SuiWatcher.test.ts
@@ -1,34 +1,101 @@
-import { expect, jest, test } from '@jest/globals';
+import { beforeAll, expect, jest, test } from '@jest/globals';
 import { INITIAL_DEPLOYMENT_BLOCK_BY_NETWORK_AND_CHAIN } from '@wormhole-foundation/wormhole-monitor-common/dist/consts';
 import { SuiWatcher } from '../SuiWatcher';
 
 jest.setTimeout(60000);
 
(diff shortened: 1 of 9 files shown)


watcher/src/watchers/__tests__/SuiWatcher.test.ts
```diff
@@ -1,34 +1,101 @@
...
-test('getFinalizedSequenceNumber', async () => {
-  const watcher = new SuiWatcher('Mainnet');
-  const blockNumber = await watcher.getFinalizedBlockNumber();
-  console.log('Received blockNumber:', blockNumber);
-  expect(blockNumber).toBeGreaterThan(INITAL_SEQUENCE_NUMBER);
+// Public Sui fullnodes prune historical checkpoint data; tests use a dynamically-computed
+// recent range so they remain valid as the chain grows. A small head buffer avoids any
+// finality/indexer-lag edge cases.
+const HEAD_BUFFER = 100;
+const RANGE_WIDTH = 5000; // ~25 minutes of Sui checkpoints; wide enough to virtually guarantee Wormhole activity
```
evan-gray, line 15: this may not always pass, depending on future activity, but it seems okay for the time being and we can always tweak as needed.

evan-gray (approved):
seems like a reasonable start.

